PR #3, opened 2023-08-31
Set up plugin boilerplate & placeholder pages

ohltyler:
### Description

- Set up plugin boilerplate with help from https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/packages/osd-plugin-generator
- Added basic plugin structure with some placeholder pages
- Added basic URL paths and breadrumbs

What it looks like:
[screen-capture (10).webm](https://github.com/opensearch-project/opensearch-ai-flow-dashboards/assets/62119629/e9046ea4-9b0f-4c68-8a19-3985b6d5541b)

### Check List


package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "ai-flow-dashboards",
```
joshpalis, line 2: shouldn't this package name be `opensearch-ai-flow-dashboards`?

package.json
```diff
@@ -0,0 +1,28 @@
...
+  "version": "3.0.0.0",
+  "description": "OpenSearch AI Flow Dashboards Plugin",
+  "main": "index.js",
+  "config": {
+    "plugin_version": "3.0.0.0",
+    "plugin_name": "aiFlowDashboards",
+    "plugin_zip_name": "ai-flow-dashboards"
```
joshpalis, line 9: same question for these plugin/zip names

package.json
```diff
@@ -0,0 +1,28 @@
...
+  },
+  "private": true,
+  "scripts": {
+    "plugin-helpers": "../../scripts/use_node ../../scripts/plugin_helpers",
+    "osd": "../../scripts/use_node ../../scripts/osd",
+    "opensearch": "node ../../scripts/opensearch",
+    "lint": "node ../../scripts/eslint .",
```
BSFishy, line 16: Why do some of these use `use_node` and others don't?

Also, plugins should also have a `lint:style` script, that calls `../../scripts/use_node ../../scripts/stylelint` (which should be part of the CI workflow, along with `lint:es`)

package.json
ohltyler: I'm following the standard naming conventions of OSD plugins, which is `<plugin-name>` + `'dashboards'` - may be camel case or kebab case

package.json
ohltyler: See https://github.com/opensearch-project/opensearch-ai-flow-dashboards/pull/3#discussion_r1316201499

joshpalis (approved):

package.json
ohltyler: I'll clean up the `use_node` vs. `node` differences, I referenced some of this from different existing plugins.

Thanks for the suggestions on the lint scripts - I'll work on that + adding a github workflow for them

public/utils/index.ts
```diff
@@ -0,0 +1,6 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export * from './constants';
```
owaiskazi19, line 6: Can't we export this directly in `constants.ts`?

owaiskazi19 (commented):
I'm very new to reviewing frontend PRs. Feel free to ignore my comment if not relevant.

ohltyler:
> I'm very new to reviewing frontend PRs. Feel free to ignore my comment if not relevant. Also, getting permission denied while viewing the screen capture.

Weird, it was working originally but now I'm seeing the same permission denied. Just re-uploaded and it's available again

public/app.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { CoreStart } from '../../../src/core/public';
+import { Navigation, APP_PATH } from './utils';
+import { Overview, UseCases, Workflows } from './pages';
+import { CoreServicesConsumer } from './core_services';
+
+interface Props extends RouteComponentProps {}
+
+export const AiFlowDashboardsApp = (props: Props) => {
+  const sidebar = (
+    <EuiPageSideBar style={{ minWidth: 190 }} hidden={false}>
```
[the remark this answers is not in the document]
ohltyler: From what I've seen typically yes, but it is also reasonable to include in-line for styling that is only used once, and for one component/element. Otherwise you would be adding too much boilerplate where a single css property would mean creating new css file, importing css file, and still having to reference the classname property for the component

public/index.ts
```diff
@@ -0,0 +1,20 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import './index.scss';
+
+import { AiFlowDashboardsPlugin } from './plugin';
+
+// This exports static code and TypeScript types,
+// as well as, OpenSearch Dashboards Platform `plugin()` initializer.
+export function plugin() {
```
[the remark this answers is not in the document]
ohltyler: It is equivalent, but from styling in OSD it follows the convention I've used

public/plugin.ts
```diff
@@ -0,0 +1,46 @@
...
+    core.application.register({
+      id: PLUGIN_ID,
+      title: 'AI Application Builder',
+      category: {
+        id: 'opensearch',
+        label: 'OpenSearch plugins',
+        order: 2000,
+      },
+      // TODO: can i remove this below order
+      order: 5000,
```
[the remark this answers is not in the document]
ohltyler: It's used by OSD to determine where this plugin is shown in the list of available plugins. Until this is finalized later on from product side, leaving a placeholder value for now

public/utils/index.ts
ohltyler: Exporting here means that the exported modules in constants are importable via `utils/` itself, rather than having to import from `utils/constants`

Main purpose of these `index.ts` files are for organizing and bubbling up exported modules so they can be imported from a higher level (`import x from ../utils/` vs. `import x from '../utils/constants`

owaiskazi19:
> Weird, it was working originally but now I'm seeing the same permission denied. Just re-uploaded and it's available again

Do we need to install some github extension to view the video because I'm seeing this:

Screenshot 2023-09-05 at 12 59 09 PM

public/plugin.ts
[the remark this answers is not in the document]
owaiskazi19: What's the order which is on line 28 then?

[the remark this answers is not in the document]
ohltyler: That is for the ordering within the "OpenSearch" plugins sub-panel. Same thing where that will be finalized later. We may not even need this ordering on line 31 later on

ohltyler:
@BSFishy I believe the issue is that my forked repo was forked before the repo was public. It is not showing as an existing fork on the repo now.

I will have a follow up PR with the [added commits](https://github.com/ohltyler/opensearch-ai-flow-dashboards/commits/plugin-bootstrap) in my [new forked repo](https://github.com/ohltyler/opensearch-ai-flow-dashboards-1)

PR #25, opened 2023-09-05
Add lint script, workflow, and pre-commit hook

ohltyler:
### Description

Follow-up to #3 

- adds a linting workflow to run lint on all src files on PR and push
- adds `precommit` npm package and configuration to automatically run the linter on every commit

### Issues Resolved

[List any issues this PR will resolve]

### Check List

  - [ ] All tests pass


.github/workflows/lint.yml
```diff
@@ -0,0 +1,48 @@
+name: Linting workflow
+on:
+  push:
+    branches:
```
owaiskazi19, line 4: I don't think so we need to specify `branches` when you want to run the workflow on all the branches anyway. I would still suggest to test the workflow on your fork and remove this line and 7th after that.

ohltyler: Thanks for the suggestion. Updated and it is running as expected still.

public/app.tsx
```diff
@@ -12,6 +12,7 @@ import { Navigation, APP_PATH } from './utils';
 import { Overview, UseCases, Workflows } from './pages';
 import { CoreServicesConsumer } from './core_services';
 
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
```
owaiskazi19, line 15: For my understanding: Is this comment needed to disable eslint? If so, why are disabling it here?

ohltyler: Its just to disable the next line which has an empty interface. Regularly the linting would flag this as an issue, but in this case it's expected.

common practice used in OSD which I'm following here is all

owaiskazi19 (approved):

Issue #4, opened 2023-09-05
Define UI component types


ohltyler:
Building out a library of hierarchical interfaces for the different component types needed for initial few use cases (semantic search / RAG). Defining styling & functionality for each. These will be based on supported steps in the use case templates.

PR #32, opened 2023-09-20
Add initial component interfaces & basic node rendering

ohltyler:
### Description

This PR accomplishes 3 main things:
1. Sets up the base interfaces that will represent all of the different building blocks / UI components users will select to construct workflows.
2. Implements interfaces for 2 components: 1/ a text embedding processor, and 2/ a k-NN index.
3. Sets up a basic `WorkflowComponent` React component that's purpose is to take in a component instance (a particular k-NN index, for example), and dynamically render based on its configuration. An example with more details is shown below.

Example:
- shows the same `WorkflowComponent` building out instances of a text embedding processor and a k-NN index.
- k-NN index component supports creation, so it has a tabbed portion to toggle the inputs based on if user wants to create a new index, or use existing
- based on the input types, renders differently (text vs. select vs. JSON, etc.)
- inputs/outputs displayed there for readability & debugging purposes
- dummy values used for the indices and the created components

[screen-capture (15).webm](https://github.com/opensearch-project/opensearch-ai-flow-dashboards/assets/62119629/5200eacb-e792-4de9-a8d7-7b8bb87d9b7b)

Adding a `rapid` label to indicate this PR may not be perfectly clean or have tests. It is for rapid iteration.

### Issues Resolved

Makes progress on #4. 

### Check List

### DIFF
--- a/public/component_types/base_interfaces.ts
+++ b/public/component_types/base_interfaces.ts
@@ -1,0 +1,86 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { COMPONENT_CATEGORY } from '../utils';
+
+/**
+ * ************ Types **************************
+ */
+
+// TODO: may change some/all of these to enums later
+export type BaseClass = string;
+export type UIFlow = string;
+export type FieldType = 'string' | 'json' | 'select';
+
+/**
+ * ************ Base interfaces ****************
+ */
+
+/**
+ * Represents a single base class as an input handle for a component.
+ * It may be optional. It may also accept multiples of that class.
+ */
+export interface IComponentInput {
+  id: string;
+  label: string;
+  baseClass: string;
+  optional: boolean;
+  acceptMultiple: boolean;
+}
+
+/**
+ * An input field for a component. Specifies enough configuration for the
+ * UI node to render it properly within the component (show it as optional,
+ * put it in advanced settings, placeholder values, etc.)
+ */
+export interface IComponentField {
+  label: string;
+  type: FieldType;
+  placeholder?: string;
+  optional?: boolean;
+  advanced?: boolean;
+}
+
+/**
+ * Represents the list of base classes as a single output handle for
+ * a component.
+ */
+export interface IComponentOutput {
+  id: string;
+  label: string;
+  baseClasses: BaseClass[];
+}
+
+/**
+ * The base interface the components will implement.
+ */
+export interface IComponent {
+  id: string;
+  type: BaseClass;
+  label: string;
+  description: string;
+  // will be used for grouping together in the drag-and-drop component library
+  category: COMPONENT_CATEGORY;
+  // determines if this component allows for new creation. this means to
+  // allow a "create" option on the UI component, as well as potentially
+  // include in the use case template construction ('provisioning' flow)
+  allowsCreation: boolean;
+  // determines if this is something that will be included in the use
+  // case template construction (query or ingest flows). provisioning flow
+  // is handled by the allowsCreation flag above.
+  isApplicationStep: boolean;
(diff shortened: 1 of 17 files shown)


public/component_types/base_interfaces.ts
```diff
@@ -0,0 +1,86 @@
...
+import { COMPONENT_CATEGORY } from '../utils';
+
+/**
+ * ************ Types **************************
+ */
+
+// TODO: may change some/all of these to enums later
+export type BaseClass = string;
+export type UIFlow = string;
+export type FieldType = 'string' | 'json' | 'select';
```
jackiehanyang, line 15: are we going to add more field type later? if so should we use an enum for this?

public/component_types/base_interfaces.ts
```diff
@@ -0,0 +1,86 @@
...
+  // the list of base classes that will be used in the component output
+  baseClasses?: BaseClass[];
+  inputs?: IComponentInput[];
+  fields?: IComponentField[];
+  // if the component supports creation, we will have a different set of input fields
+  // the user needs to fill out
+  createFields?: IComponentField[];
+  outputs?: IComponentOutput[];
+  // we will need some init function when the component is drug into the workspace
+  init?(): Promise<any>;
```
jackiehanyang, line 85: should we use `Promise<T>` where T is the type of data returned for clarity?

public/component_types/base_interfaces.ts
ohltyler: Most likely will - see comment on line 12

public/component_types/base_interfaces.ts
ohltyler: This signature will likely change a lot. Prefer to leave loosely for now, in the end it will be typesafe.

public/pages/workflow_builder/components/input_field_list.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export function InputFieldList(props: InputFieldListProps) {
+  return (
+    <EuiFlexItem>
+      {props.inputFields?.map((field, idx) => {
+        let el;
+        switch (field.type) {
+          case 'string': {
+            el = (
+              <EuiFlexItem key={idx}>
```
jackiehanyang, line 28: using index as a key isn't always a problem. But if the list can change over time, will it cause issues and even potentially introduce bugs? Do we have other unique identifier in `IComponentField` that we can use and the key here?

public/pages/workflow_builder/components/input_field_list.tsx
```diff
@@ -0,0 +1,62 @@
...
+interface InputFieldListProps {
+  inputFields?: IComponentField[];
+}
+
+export function InputFieldList(props: InputFieldListProps) {
+  return (
+    <EuiFlexItem>
+      {props.inputFields?.map((field, idx) => {
+        let el;
+        switch (field.type) {
```
jackiehanyang, line 25: nit: instead of using a switch statement, consider refactoring it to use a mapping object for better clarity and scalability. This also makes adding new input types in the future easier

public/pages/workflow_builder/components/input_field_list.tsx
ohltyler: Using the index as a key for items in a list is a common JS pattern since it is calculated on-the-fly for each render. This will likely change a lot as we iterate so prefer to leave for now.

public/pages/workflow_builder/components/input_field_list.tsx
ohltyler: Thanks. Likely this will be moved into more complex helper fns as the number of object types will grow, and the complexity to render them will grow. Prefer to just leave as a simple switch for now.

jackiehanyang (approved):

Issue #10, opened 2023-09-05
Read-only workspace


ohltyler:
Build out the workspace area to construct the workflows. Initially we may have the workspace be read-only regarding the components and connections between them. We will only allow users to input the required fields within the components. We will also want to show validation and any errors of a given workflow. This is pending final design from @opensearch-project/opensearch-ux

ohltyler:
Refactoring into clearer meta issues.

Issue #8, opened 2023-09-05
Basic workflow validation and conversion


ohltyler:
Support basic validation and conversion of a given UI flow into a use case template. Can start with just supporting a few static use cases.

ohltyler:
Refactoring into clearer meta issues.

PR #33, opened 2023-09-21
Set up redux infrastructure

ohltyler:
### Description

Sets up the initial redux infrastructure for maintaining frontend state. We choose to use redux / react-redux / redux-toolkit libraries since it is the consistent way that core OSD + other plugins use, and is the standard across modern react applications. It also doesn't add any dependencies since we consume them from core OSD.
- sets up global redux store via `configureStore`
- sets up basic workspace reducer to process workspace-related state updates via `createSlice`
- wraps the entire app within a `Provider` so React components have access to the store
- sets up sample cmds for future reference as ways to access and set the store in the base `WorkflowBuilder` page

Tested with some basic functionalities, including adding some dummy components, and validating if there are changes. These are using the `components` and `isDirty` vars within the `workspace` state.

[screen-capture (16).webm](https://github.com/opensearch-project/opensearch-ai-flow-dashboards/assets/62119629/56c1b67d-4251-405f-9eb0-6feefee54676)

### Issues Resolved

Makes progress on #8 & #10

### Check List

### DIFF
--- a/public/pages/workflow_builder/workflow_builder.tsx
+++ b/public/pages/workflow_builder/workflow_builder.tsx
@@ -4,6 +4,7 @@
  */
 
 import React, { useEffect } from 'react';
+// import { useSelector, useDispatch } from 'react-redux';
 import {
   EuiPageHeader,
   EuiFlexGroup,
@@ -14,8 +15,15 @@ import {
 } from '@elastic/eui';
 import { BREADCRUMBS } from '../../utils';
 import { getCore } from '../../services';
+// import { AppState, removeDirty, setComponents } from '../../store';
 
 export function WorkflowBuilder() {
+  // TODO: below commented out lines can be used for fetching & setting global redux state
+  // const dispatch = useDispatch();
+  // const { isDirty, components } = useSelector(
+  //   (state: AppState) => state.workspace
+  // );
+
   useEffect(() => {
     getCore().chrome.setBreadcrumbs([
       BREADCRUMBS.AI_APPLICATION_BUILDER,
@@ -36,7 +44,7 @@ export function WorkflowBuilder() {
       </EuiPageHeader>
       <EuiSpacer size="l" />
       <EuiFlexItem>
-        <EuiText>Placeholder for workflow builder page...</EuiText>
+        <EuiText>Workspace</EuiText>
       </EuiFlexItem>
     </div>
   );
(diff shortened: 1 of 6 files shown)


public/store/reducers/workspace_reducer.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+import { createSlice } from '@reduxjs/toolkit';
+
+const initialState = {
+  isDirty: false,
+  components: [],
+};
+
+const workspaceSlice = createSlice({
```
owaiskazi19, line 13: Silly questions: But don't we have to treat this as an actionType?

public/store/reducers/workspace_reducer.ts
```diff
@@ -0,0 +1,31 @@
...
+  name: 'workspace',
+  initialState,
+  reducers: {
+    setDirty(state) {
+      state.isDirty = true;
+    },
+    removeDirty(state) {
+      state.isDirty = false;
+    },
```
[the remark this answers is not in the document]
ohltyler: We can - prefer to leave for now and as more implementation comes we can decide on if we want a pattern like this, or what you're suggesting. It may get complex where one action may be updating several different state vars.

public/store/reducers/workspace_reducer.ts
ohltyler: `createSlice` takes care of that for us, where we don't need to create action strings at all. See https://redux.js.org/tutorials/fundamentals/part-8-modern-redux#writing-slices

public/pages/workflow_builder/workflow_builder.tsx
```diff
@@ -14,8 +15,15 @@ import {
 } from '@elastic/eui';
 import { BREADCRUMBS } from '../../utils';
 import { getCore } from '../../services';
+// import { AppState, removeDirty, setComponents } from '../../store';
 
 export function WorkflowBuilder() {
+  // TODO: below commented out lines can be used for fetching & setting global redux state
```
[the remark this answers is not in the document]
ohltyler: I don't have an explicit issue for the redux state, kind of treating it implicitly in #8 and #10. The commented out lines show how to use dispatch to call on the `setDirty`/`removeDirty`/ etc. action creators.

public/store/reducers/workspace_reducer.ts
owaiskazi19: Oh! Something new for me. I'll read about it.

owaiskazi19 (approved):
Looks good on my end but I'll wait for @jackiehanyang review before merging it.

jackiehanyang:
LGTM!

jackiehanyang (approved):

Issue #6, opened 2023-09-05
[META] Server-side logic


ohltyler:
We will need several different iterations and additions to the server-side API calls to OpenSearch:

ohltyler:
Refactoring into clearer meta issues.

PR #37, opened 2023-09-22
Add redux thunk fns; add cat indices API

ohltyler:
### Description

This PR adds support for e2e processing of fetching visible indices. It does this by:
1. Adding a server-side `cat.indices` route, a client-side fn in the `RouteService`, and a redux thunk fn in a new `opensearchSlice` reducer. This follows the modern pattern of async / server-side data handling with redux. For more details, see https://redux.js.org/tutorials/fundamentals/part-8-modern-redux#writing-thunks

### Issues Resolved

Makes progress on #6, #10

### Check List

### DIFF
--- a/common/constants.ts
+++ b/common/constants.ts
@@ -5,4 +5,6 @@
 export const PLUGIN_ID = 'aiFlowDashboards';
 
 export const BASE_NODE_API_PATH = '/api/ai_flow';
-export const SEARCH_PATH = `${BASE_NODE_API_PATH}/search`;
+export const BASE_INDICES_NODE_API_PATH = `${BASE_NODE_API_PATH}/indices`;
+export const SEARCH_INDICES_PATH = `${BASE_INDICES_NODE_API_PATH}/search`;
+export const FETCH_INDICES_PATH = `${BASE_INDICES_NODE_API_PATH}/fetch`;
--- a/public/store/reducers/opensearch_reducer.ts
+++ b/public/store/reducers/opensearch_reducer.ts
@@ -1,0 +1,58 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { createAsyncThunk, createSlice } from '@reduxjs/toolkit';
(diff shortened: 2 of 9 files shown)


common/constants.ts
```diff
@@ -5,4 +5,6 @@
 export const PLUGIN_ID = 'aiFlowDashboards';
 
 export const BASE_NODE_API_PATH = '/api/ai_flow';
```
owaiskazi19, line 7: Why do we have `api` in the base path?

public/store/reducers/opensearch_reducer.ts
```diff
@@ -0,0 +1,53 @@
...
+  name: OPENSEARCH_PREFIX,
+  initialState,
+  reducers: {},
+  extraReducers: (builder) => {
+    builder
+      .addCase(fetchIndices.pending, (state, action) => {
+        state.loading = true;
+      })
+      .addCase(fetchIndices.fulfilled, (state, action) => {
+        const indicesMap = new Map<string, IIndex>();
```
owaiskazi19, line 39: Should we declare the map on top of createSlice on line 28?

owaiskazi19 (commented):
Looks good overall with small comments/questions

common/constants.ts
ohltyler: It follows the standard node API naming conventions. Typically `api/<plugin>/<do-something>`. This way frontend plugins won't have name clashing when designing their node APIs

owaiskazi19: Ohh it's for Dashboard plugin. I see.

public/store/reducers/opensearch_reducer.ts
ohltyler: That was my first attempt :) but it isn't serializable so I converted to a plain JS object

ohltyler: ref: https://stackoverflow.com/questions/67763839/why-is-a-map-in-js-considered-non-serializable-in-redux

owaiskazi19 (approved):

Issue #12, opened 2023-09-05
Persisting UI metadata


ohltyler:
Design & implement how to persist the ReactFlow JSON somewhere in the backend (part of use-case template, separate system index, etc.). May need some backend implementation.

PR #42, opened 2023-09-25
Add ReactFlow to Workflow Details page

ohltyler:
### Description

This PR adds a new ReactFlow instance into the workflow details page.

More details:
- Uses the ReactFlow documentation for persisting ReactFlow state (nodes / edges) in a React component via `useNodesState()` / `useEdgesState()`, respectively. It also includes defining other flow lifecycle functions, like `onDrop()`, `onDragOver()`, and `onConnect()`. This is where we will define how we want to manipulate the ReactFlow state, and gives us the opportunity for any other data we may need to fetch from redux store (or trigger any API calls ourselves)
- Adds a ReactFlowContextProvider in order to persist ReactFlow state. This provider wraps the entire application so all nested components can consume and access contextual values via `useContext()`. By default, we add a few stub functions to delete nodes and edges, which are not something that comes by default via `<ReactFlow>` component. We don't add this in redux store since it serves a different purpose (sharing mostly-static data across components vs. having a centralized datastore that is updated often).
- The workflow state and the low-level ReactFlow JSON I have hardcoded for now; it is still TBD on where this information will live exactly, or how much we want/need to generate on-the-fly.

Next steps:
- Integrate the added components in #32 as ReactFlow node types, such that they are visible in the ReactFlow workspace
- Improve the workspace state & datastore. Can use hardcoded values for now, but should have the following state for each workflow: 1/ high-level redux store persisting the workspace state, and 2/ low-level ReactFlow JSON. This way, each workflow can be rendered and show a unique workflow state when loading the details page. The exact difference between 1/ and 2/ is still TBD

Demo video w/ hardcoded workspace state:

[screen-capture (18).webm](https://github.com/opensearch-project/opensearch-ai-flow-dashboards/assets/62119629/830355d4-2e6f-40f2-b822-ad90d52d83ea)

### Issues Resolved

Makes progress on #10, #12

### Check List


amitgalitz:
nit: should we use the newest reactflow version available right now? 11.9.2

amitgalitz:
> nit: should we use the newest reactflow version available right now? 11.9.2

Oh I just realized 11.8.3 -> 11.9.x happened 12 hours ago, just looked at the release date. Anyways we can probably wait as I see they had fast iteration with 11.9.1/2 fixing things in 11.9.0

ohltyler:
> > nit: should we use the newest reactflow version available right now? 11.9.2
> 
> Oh I just realized 11.8.3 -> 11.9.x happened 12 hours ago, just looked at the release date. Anyways we can probably wait as I see they had fast iteration with 11.9.1/2 fixing things in 11.9.0

Also, it will be changing to `xyflow` sometime in the near future: https://wbkd.notion.site/Upcoming-Changes-at-React-Flow-1a443641891a4069927c0a115e915251. Depending on when that happens, we may update to that. The earlier, the better.

common/helpers.ts
```diff
@@ -0,0 +1,74 @@
...
+      string | undefined
+    >
+  >;
+
+  const dummyEdges = [
+    {
+      id: 'e1-2',
+      source: 'model',
+      target: 'ingest-pipeline',
+      style: {
```
amitgalitz, line 58: I know this is going to change, but just to move conversation from offline to the github PR here. Want to confirm that we might potentially either have two different places or one from where this data will come form. One is through a redux call to an index where we have saved workflows with corresponding UI metadata. The other could be if we decide to go that way a library with fixed templates to pre-populate the node.

public/pages/workflow_detail/workspace/workspace.tsx
```diff
@@ -3,24 +3,161 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import React from 'react';
+import React, { useRef, useContext, useCallback, useEffect } from 'react';
```
amitgalitz, line 6: nit: unrelated to this file but we should use prettier like we do in AD https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/DEVELOPER_GUIDE.md#formatting.

We should probably try and stick to some specific formatter from day one. I can open an issue if we don't want to do it in this PR

public/pages/workflow_detail/workspace/workspace.tsx
```diff
@@ -3,24 +3,161 @@
...
+        id: Date.now().toFixed(),
+        type: nodeData.type,
+        position,
+        data: { label: nodeData.label },
+        style: {
+          background: 'white',
+        },
+      };
+
+      // TODO: add logic to add node into the redux datastore
```
amitgalitz, line 89: I thought we aren't storing the nodes in the redux while updating the workflow and just the hook here? Is that in-correct. We will update a redux datastore as we drop in nodes?

public/pages/workflow_detail/workspace/workspace.tsx
```diff
@@ -3,24 +3,161 @@
...
+    // connection details.
+    [setEdges]
+  );
+
+  const onDragOver = useCallback((event) => {
+    event.preventDefault();
+    event.dataTransfer.dropEffect = 'move';
+  }, []);
+
+  const onDrop = useCallback(
```
amitgalitz, line 55: Might require its own different conversation but for things like credentials that users might not want to type in here on the UI we might need to pre-populate some variable during "on drop" here if we allow some pre-configuration

public/pages/workflow_detail/workspace/workspace.tsx
```diff
@@ -3,24 +3,161 @@
...
+
+      setNodes((nds) => nds.concat(newNode));
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [reactFlowInstance]
+  );
+
+  // Initialization hook
+  useEffect(() => {
+    // TODO: fetch the nodes/edges dynamically (loading existing flow,
```
amitgalitz, line 99: Will all these options live here? So basically we will have some check if we have some current flowID in our state? or from our in house fixed templates and etc.

public/pages/workflow_detail/workspace/workspace.tsx
```diff
@@ -3,24 +3,161 @@
...
+      if (typeof nodeData === 'undefined' || !nodeData) {
+        return;
+      }
+
+      // Fetch bounds based on the ref'd div component, adjust as needed.
+      // TODO: remove hardcoded bounds and fetch from a constant somewhere
+      // @ts-ignore
+      const reactFlowBounds = reactFlowWrapper.current.getBoundingClientRect();
+      // @ts-ignore
+      const position = reactFlowInstance.project({
```
amitgalitz, line 71: just wondering, where does the height and width info come from for these nodes to be rendered?

public/store/context/react_flow_context_provider.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * values are set so any nested components can use useContext to access these
+ * values.
+ *
+ * This is how we can manage ReactFlow context consistently across the various
+ * nested child components.
+ */
+export function ReactFlowContextProvider({ children }: any) {
+  const [reactFlowInstance, setReactFlowInstance] = useState(null);
+
+  const deleteNode = (nodeId: string) => {
```
amitgalitz, line 28: Why isn't this where the `onConnect` `onDrop` and other workspace functions are. I understand why it can live here also but surprised its not all in the same place as other functions

common/helpers.ts
ohltyler: Right, yes. The styling of different component types and edges will be standardized as we iterate and see what sort of base data we need for all component and edge types, and what separate fields we need to differ between them.

public/pages/workflow_detail/workspace/workspace.tsx
ohltyler: I've integrated with prettier on the initial commit #2. I've not run into any auto-formatting issues, are you having issues on your local maybe?

public/pages/workflow_detail/workspace/workspace.tsx
ohltyler: Thanks for the catch, yes you're correct. We will utilize reactFlowContext and useNodesState() / useEdgesState() to maintain all of this for us. I'll remove the comment

public/pages/workflow_detail/workspace/workspace.tsx
ohltyler: Yeah, we can handle this logic outside of the overall workspace component and treat it at the node/component-level. Flowise for example, has a CredentialInputHandler which is a special handler rendered when a credential input is included in the component's interface.

This will require some more digging to understand and implement a secure way for this part, though. Let me add an issue to track this, thanks for the callout.

public/pages/workflow_detail/workspace/workspace.tsx
ohltyler: We will need to have some initialization logic to determine whether the page is loaded from an existing workflow, a new workflow, or (potentially) redirected from a sample template as a starting point, similar to Flowise.

For loading an existing workflow, we can parse the url for a workflow ID, fetch workflow details that contain the set of nodes/edges in the UI metadata, and load them into the reactflow instance via setNodes() / setEdges().

For starting with a blank canvas, we can simply set nodes/edges to empty arrays.

For redirecting from a sample template, we can focus on that as the UX design becomes more clear on that flow and if it is something we plan to do.

public/pages/workflow_detail/workspace/workspace.tsx
ohltyler: The `reactFlowWrapper` contains the element positioning in the webpage, and is used for calculating the bounds on line 69. The `position` based on these hardcoded bounds is saying that for every element that is dragged into the workspace, place it 80 pixels from the left bounds, and 90 pixels from the top bounds. We will likely want this to be dynamic based on the user mouse, for example, but just placing those hardcoded values for ease of use for now.

public/store/context/react_flow_context_provider.tsx
ohltyler: It is part of the context so the functions will be easily accessible by other React components later on that are not directly stored in the main `Workspace` component (and may not even be children of `Workspace`, such as helper menus to delete the component, or handlers for removing connections of edges of a UI component, etc.

The `onConnect`/`onDrop`/etc. fns are included just in the `Workspace` for now, but may potentially be refactored or added into the context later on if that pattern will make sense. Right now, the configuration I have matches with the Flowise pattern which is beneficial for us to start with for now and change as needed down the line.

amitgalitz (approved):

dbwiddis (approved):

Issue #16, opened 2023-09-05
Support workflow state


ohltyler:
Have a loading bar or somewhere to view workflow status after a user submits/creates a new workflow. May involve new API integrations to process some of the state from the backend framework. May be updated as the design is finalized from @opensearch-project/opensearch-ux

ohltyler:
Refactoring into clearer meta issues.

PR #49, opened 2023-10-03
Add input and output handles to components

ohltyler:
### Description

This PR adds the input and output handles to the general `WorkspaceComponent`. Specifically, we add an `InputHandle` and `OutputHandle` class which adds the `Handle` reactflow component dynamically based on the component spec. These classes configure the data within the input and output handles so they can be processed correctly.

Also adds a custom `isValidConnection` helper fn to determine if a connection from a source to target is allowed. Currently, we do basic validation that 1/ the base classes are the same and 2/ prevent multiple connections if the input handle does not allow it. Short video below demonstrates this.

Current logic assumes the output handle may be multiple classes (string[]). These are serialized/deserialized by turning them into a pipe-delimited string. The input handle is under the assumption to just have one class for now.

Additional clean up
- removed unnecessary `id` field from the component classes. The unique `id` for a node will just use the existing reactflow node that already has an `id`. When a user adds a component to the workspace (`onDrop()`), a random ID will be generated using the new `generateId` helper fn.
- formalizes the classes from type `string` to a `COMPONENT_CLASS` enum

Testing
- tested the `isValidConnection` fn works as expected for the different use cases
- tested that 0/1/multiple input & output handles renders correctly

[screen-capture (12).webm](https://github.com/opensearch-project/opensearch-ai-flow-dashboards/assets/62119629/8017c349-28a6-4c0f-b65c-928a096c8221)

### Issues Resolved

Makes progress on #8, #10, and #16

### Check List

### DIFF
--- a/public/component_types/indices/knn_index.ts
+++ b/public/component_types/indices/knn_index.ts
@@ -44,7 +44,15 @@ export class KnnIndex implements IComponent {
     // that will be referenced/used as input across multiple flows
     this.allowedFlows = ['Ingest', 'Query', 'Other'];
     this.baseClasses = [this.type];
-    this.inputs = [];
+    this.inputs = [
+      {
+        id: 'text-embedding-processor',
+        label: 'Text embedding processor',
+        baseClass: 'text_embedding_processor',
+        optional: false,
+        acceptMultiple: false,
+      },
+    ];
     this.fields = [
       {
         label: 'Index Name',
--- a/public/component_types/indices/knn_index.ts
+++ b/public/component_types/indices/knn_index.ts
@@ -3,37 +3,34 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { COMPONENT_CATEGORY } from '../../utils';
+import { COMPONENT_CATEGORY, COMPONENT_CLASS } from '../../utils';
 import {
   IComponent,
   IComponentField,
   IComponentInput,
   IComponentOutput,
   UIFlow,
-  BaseClass,
 } from '../interfaces';
 
 /**
  * A k-NN index UI component
  */
 export class KnnIndex implements IComponent {
-  id: string;
-  type: BaseClass;
+  type: COMPONENT_CLASS;
   label: string;
   description: string;
   category: COMPONENT_CATEGORY;
   allowsCreation: boolean;
   isApplicationStep: boolean;
   allowedFlows: UIFlow[];
-  baseClasses: BaseClass[];
+  baseClasses: COMPONENT_CLASS[];
   inputs: IComponentInput[];
   fields: IComponentField[];
   createFields: IComponentField[];
   outputs: IComponentOutput[];
 
   constructor() {
-    this.id = 'knn_index';
-    this.type = 'knn_index';
+    this.type = COMPONENT_CLASS.KNN_INDEX;
     this.label = 'k-NN Index';
     this.description = 'A k-NN Index to be used as a vector store';
     this.category = COMPONENT_CATEGORY.INDICES;
@@ -48,7 +45,7 @@ export class KnnIndex implements IComponent {
       {
         id: 'text-embedding-processor',
         label: 'Text embedding processor',
-        baseClass: 'text_embedding_processor',
+        baseClass: COMPONENT_CLASS.TEXT_EMBEDDING_PROCESSOR,
         optional: false,
         acceptMultiple: false,
       },
@@ -82,7 +79,6 @@ export class KnnIndex implements IComponent {
     ];
     this.outputs = [
       {
-        id: this.id,
         label: this.label,
         baseClasses: this.baseClasses,
       },
--- a/public/pages/workflow_detail/workspace_component/utils.ts
+++ b/public/pages/workflow_detail/workspace_component/utils.ts
@@ -1,0 +1,58 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Connection, ReactFlowInstance } from 'reactflow';
+import { IComponentInput } from '../../../../common';
+
+/**
+ * Collection of utility functions for the workspace component
+ */
+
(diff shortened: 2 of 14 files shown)


dbwiddis (approved):

public/component_types/indices/knn_index.ts
```diff
@@ -3,37 +3,34 @@
...
-  type: BaseClass;
+  type: COMPONENT_CLASS;
   label: string;
   description: string;
   category: COMPONENT_CATEGORY;
   allowsCreation: boolean;
   isApplicationStep: boolean;
   allowedFlows: UIFlow[];
-  baseClasses: BaseClass[];
+  baseClasses: COMPONENT_CLASS[];
```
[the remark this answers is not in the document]
ohltyler: Yep. These component classes don't mean anything right now until it is more finalized what the UX building blocks will look like.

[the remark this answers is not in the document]
amitgalitz: Actually 10 seconds later I see we have `this.category = COMPONENT_CATEGORY.INDICES;` is the decided pattern for each component to have a base class and category? It wasn't 100% clear to follow what category should be and what class should be and how we standardize the nesting.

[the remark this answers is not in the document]
ohltyler: All the implementations right now are just to POC the UI works as expected.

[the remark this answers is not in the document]
ohltyler: We don't know the decided pattern yet and how the parent/child relationships will look like just yet.

public/component_types/indices/knn_index.ts
```diff
@@ -44,7 +41,15 @@ export class KnnIndex implements IComponent {
     // that will be referenced/used as input across multiple flows
     this.allowedFlows = ['Ingest', 'Query', 'Other'];
     this.baseClasses = [this.type];
-    this.inputs = [];
+    this.inputs = [
+      {
+        id: 'text-embedding-processor',
+        label: 'Text embedding processor',
+        baseClass: COMPONENT_CLASS.TEXT_EMBEDDING_PROCESSOR,
```
amitgalitz, line 48: Again, aware a lot of the questions I am asking might be not relevant as lots hasn't been decided. Here do we have the processor as an input for the index?  I thought we wont be doing that as the processor is an input to the ingest pipeline creation.

ohltyler: Yes it's undecided, refer to above comments.

public/utils/utils.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// Append 16 random characters
+export function generateId(prefix: string) {
```
amitgalitz, line 7: just wondering is something like this used in other places. I know in backend we have some auto ID generator methods or using the document ID

ohltyler: I haven't come across anything on OSD side but I'll check with couple of them.

public/pages/workflow_detail/workspace_component/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+    return ref.current.offsetTop + ref.current.clientHeight / 2;
+  } else {
+    return 0;
+  }
+}
+
+// Validates that connections can only be made when the source and target classes align, and
+// that multiple connections to the same target handle are not allowed unless the input configuration
+// for that particular component allows for it.
+export function isValidConnection(
```
amitgalitz, line 28: quick question if we do here the validation if the source class aligns with target class, what is left for the onConnect method where we do validation as well?

public/pages/workflow_detail/workspace_component/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+  connection: Connection,
+  rfInstance: ReactFlowInstance
+): boolean {
+  const sourceHandle = connection.sourceHandle;
+  const targetHandle = connection.targetHandle;
+  const targetNodeId = connection.target;
+  const inputClass = sourceHandle || '';
+  // We store the output classes in a pipe-delimited string. Converting back to a list.
+  const outputClasses = targetHandle?.split('|') || [];
```
amitgalitz, line 37: Did you test here if we have a list of 3 output classes, and a list of 3 target classes, just one of them matching is enough?

public/pages/workflow_detail/workspace_component/utils.ts
ohltyler: `onConnect` will handle any other updates we need to do after a connection is made, such as updating properties of the edge itself, or update properties of the node being connected potentially. `isValidConnection` is just for gatekeeping that a connection is allowed to be made or not.

public/component_types/indices/knn_index.ts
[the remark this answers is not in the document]
amitgalitz: Sounds good, I imagine if we have more fine tuned categories and component classes and types here later on it might effect validation but that can be done later

public/utils/utils.ts
amitgalitz: Does flowise create a node Id like this also?

amitgalitz (approved):

public/utils/utils.ts
ohltyler: Yes- we need them to differentiate if there are multiple components of the same type in a workspace.

ohltyler: There is nothing concrete on OSD side for this, and to prevent more coupling I think it is best we keep this util fn.

public/pages/workflow_detail/workspace_component/utils.ts
ohltyler: Yes. Also, I was splitting the target handle instead of the source handle. Updated in latest commit.

ohltyler: I got rid of input/output language since that added some confusion.

amitgalitz (approved):

Issue #18, opened 2023-09-05
Custom workflow validation & conversion


ohltyler:
Support an arbitrary flow of nodes/edges and validate it + be able to convert it to a use case template.

ohltyler:
Refactoring into clearer meta issues.

PR #51, opened 2023-10-04
Clean up obj serialization; set up save workflow stub fns

ohltyler:
### Description

This PR does a few different things:
1. Cleans up `Workflow` serialization in a few different ways: 1/ We now store the reactflow state under `workspaceFlowState` field in a `Workflow`. We do some post-processing on the nodes so we don't store unnecessary data in the state. 2/ Serialize the workspace components (knn index, text embedding processor, etc.) by converting class -> obj. To do this we add a `toObj()` fn in a base component class that all of the component classes extend. We may have future utility functions added here so this sets that up.
2. Sets up stub functions for the save workflow steps, including flow validation & use case template generation.
3. Remove `components` field from the `workspace` reducer, since we simply utilize the ReactFlowInstance for keeping this state.
4. Utilizing `setDirty()` / `removeDirty` redux fns when a node or edge is added (deletions are TBD since there is no UX design on this part yet). See demo video below, where the 'save' button is only clickable when there is a change in the workspace state (adding an edge), and gets disabled after it has been "saved":

[screen-capture (13).webm](https://github.com/opensearch-project/opensearch-ai-flow-dashboards/assets/62119629/0ef80782-cdf0-42be-9156-0bbba9aa09ea)

### Issues Resolved

Makes progress on #10, #16, #18

### Check List

### DIFF
--- a/common/interfaces.ts
+++ b/common/interfaces.ts
@@ -28,7 +28,7 @@ type ReactFlowViewport = {
   zoom: number;
 };
 
-export type ReactFlowState = {
+export type WorkspaceFlowState = {
   nodes: ReactFlowComponent[];
   edges: ReactFlowEdge[];
   viewport?: ReactFlowViewport;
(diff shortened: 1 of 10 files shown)


ohltyler:
Rebased to resolve merge conflicts

dbwiddis (approved):

PR #57, opened 2023-10-07
Add a component details side panel

ohltyler:
### Description

This PR adds a resizable panel that dynamically shows component details based on the selected component in the ReactFlow workspace, using ReactFlow's [useOnSelectionChange() hook](https://reactflow.dev/docs/api/hooks/use-on-selection-change/). If none are selected, it shows an empty state. If the panel is closed, it will automatically open when a user selects a component in the workspace. The panel content replaces what used to be in the components themselves; this is where the users will view / edit / update user-provided input to the individual components. Form validation & persisting input is still a TODO.

Other details:
- nests the entire workspace in a new `ResizableWorkspace` component to contain both the ReactFlow workspace, and the new `ComponentDetails` panel.
- nests the entire workspace in a `ReactFlowProvider` such that child components can access the various [hooks provided by ReactFlow](https://reactflow.dev/docs/api/hooks/use-react-flow/)
- cleans up some styling to maximize the space better and make component sizing and growing functionality consistent
- adds few high-level comments for the base page components

Demo video:

[screen-capture (14).webm](https://github.com/opensearch-project/opensearch-ai-flow-dashboards/assets/62119629/cf9abb7b-a1aa-480e-a7cd-6f1f029e04a4)

### Issues Resolved

Makes progress on #10 , #16 

### Check List

### DIFF
--- a/public/app.tsx
+++ b/public/app.tsx
@@ -19,7 +19,7 @@ interface Props extends RouteComponentProps {}
 
 export const AiFlowDashboardsApp = (props: Props) => {
   const sidebar = (
-    <EuiPageSideBar style={{ minWidth: 190 }} hidden={false}>
+    <EuiPageSideBar style={{ minWidth: 190 }} hidden={false} paddingSize="l">
       <EuiSideNav
         style={{ width: 190 }}
         items={[
@@ -50,7 +50,10 @@ export const AiFlowDashboardsApp = (props: Props) => {
   return (
     <EuiPageTemplate
       template="empty"
-      pageContentProps={{ paddingSize: 'm' }}
+      paddingSize="none"
+      grow={true}
+      restrictWidth={false}
+      pageContentProps={{ paddingSize: 's' }}
       pageSideBar={sidebar}
     >
       <Switch>
--- a/public/pages/overview/overview.tsx
+++ b/public/pages/overview/overview.tsx
@@ -14,6 +14,14 @@ import {
 import { BREADCRUMBS } from '../../utils';
 import { getCore } from '../../services';
 
+/**
+ * The overview page. This contains a detailed description on what
+ * this plugin offers, and links to different resources (blogs, demos,
+ * documentation, etc.)
+ *
+ * This may be hidden for the initial release until we have sufficient content
+ * such that this page adds enough utility & user value.
+ */
 export function Overview() {
   useEffect(() => {
     getCore().chrome.setBreadcrumbs([
--- a/public/pages/workflow_detail/workspace/component_details.tsx
+++ b/public/pages/workflow_detail/workspace/component_details.tsx
@@ -26,6 +26,11 @@ interface ComponentDetailsProps {
   isOpen: boolean;
 }
 
+/**
+ * A panel that will be nested in a resizable container to dynamically show
+ * the details and user-required inputs based on the selected component
+ * in the flow workspace.
+ */
 export function ComponentDetails(props: ComponentDetailsProps) {
   // TODO: use this instance to update the internal node state. ex: update field data in the selected node based
   // on user input
--- a/public/pages/workflow_detail/workspace/component_details.tsx
+++ b/public/pages/workflow_detail/workspace/component_details.tsx
@@ -44,6 +44,7 @@ export function ComponentDetails(props: ComponentDetailsProps) {
    * Hook provided by reactflow to listen on when nodes are selected / de-selected.
    * - populate panel content appropriately
    * - open the panel if a node is selected and the panel is closed
+   * - it is assumed that only one node can be selected at once
    */
   useOnSelectionChange({
     onChange: ({ nodes, edges }) => {
--- a/public/pages/workflow_detail/workspace/component_details.tsx
+++ b/public/pages/workflow_detail/workspace/component_details.tsx
@@ -21,12 +21,12 @@ import { InputFieldList } from '../workspace_component/input_field_list';
 // styling
 import './workspace-styles.scss';
 
-interface ComponentInputsProps {
+interface ComponentDetailsProps {
   onToggleChange: () => void;
   isOpen: boolean;
 }
 
-export function ComponentInputs(props: ComponentInputsProps) {
+export function ComponentDetails(props: ComponentDetailsProps) {
   // TODO: use this instance to update the internal node state. ex: update field data in the selected node based
   // on user input
   const { reactFlowInstance } = useContext(rfContext);
--- a/public/pages/workflow_detail/workspace/resizable_workspace.tsx
+++ b/public/pages/workflow_detail/workspace/resizable_workspace.tsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import React from 'react';
+import React, { useRef, useState } from 'react';
 import { ReactFlowProvider } from 'reactflow';
 import { EuiResizableContainer } from '@elastic/eui';
 import { Workflow } from '../../../../common';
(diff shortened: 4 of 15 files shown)


amitgalitz:
Some generic questions:
In AD-Dashboads we have `anomaly_detection_app` where we wrap the our `Main` component with our store. We also have the `Main` react component where we wrap the rest of our components with `CoreServicesConsumer`. Why is that different here? Is the way you have in this plugin cleaner since you moved what would be in Main to `app.tsx`?

Also shouldn't the logic in route_service.ts be a part of `routes/utils/opensearch.ts`. Not sure I understand having this interface outside here.

public/app.tsx
```diff
@@ -50,7 +50,10 @@ export const AiFlowDashboardsApp = (props: Props) => {
   return (
     <EuiPageTemplate
```
amitgalitz, line 51: nice, didn't know this existed vs <EuiPage> with more inner components

ohltyler:
> Some generic questions: In AD-Dashboads we have `anomaly_detection_app` where we wrap the our `Main` component with our store. We also have the `Main` react component where we wrap the rest of our components with `CoreServicesConsumer`. Why is that different here? Is the way you have in this plugin cleaner since you moved what would be in Main to `app.tsx`?

Yes to the latter. This follows the recommended pattern for OSD plugins. We don't wrap anything in `CoreServicesConsumer` since we are accessing core services via `services` which is set when the plugin is initialized. This follows the updated pattern for accessing services and is how core OSD plugins access other services, instead of wrapping everything in a consumer and using react context. Additionally this is what we did on the AD plugin recently for accessing services from `VisAugmenter` and other core OSD plugin services.

> 
> Also shouldn't the logic in route_service.ts be a part of `routes/utils/opensearch.ts`. Not sure I understand having this interface outside here.

This follows the pattern provided by OSD for interfacing with the server side. `route_service` lives on client side and exposes clean fns to make API calls to the server side. `routes/utils/opensearch.ts` is the server side that will forward requests to OpenSearch.

Worth noting that these helper fns in `route_service` will primarily be used in redux async thunks to integrate with the redux store ([ref](https://github.com/opensearch-project/opensearch-ai-flow-dashboards/blob/main/public/store/reducers/opensearch_reducer.ts#L19)). Then, from client-side, all we need to do is call `fetchIndices()` redux fn which will asynchronously fetch the indices & persist them in the redux store. See https://github.com/opensearch-project/opensearch-ai-flow-dashboards/pull/37 for details on that.

public/pages/overview/overview.tsx
```diff
@@ -14,6 +14,14 @@ import {
 import { BREADCRUMBS } from '../../utils';
 import { getCore } from '../../services';
 
+/**
```
amitgalitz, line 17: could be where we show sample templates

public/app.tsx
ohltyler: A lot has changed in OpenSearch & OUI since the inception of the AD plugin :)

public/pages/workflow_detail/workspace/resizable_workspace.tsx
```diff
@@ -0,0 +1,66 @@
...
+      style={{ marginLeft: '-14px' }}
+    >
+      {(EuiResizablePanel, EuiResizableButton, { togglePanel }) => {
+        if (togglePanel) {
+          collapseFn.current = (panelId: string, { direction }) =>
+            togglePanel(panelId, { direction });
+        }
+
+        return (
+          <ReactFlowProvider>
```
amitgalitz, line 44: Additional clarification question:

From my understanding a React `context` is used so the parent component make some information available to any component in the tree below it. We have our `ReactFlowContextProvider`, does this have the info on all reactflowInstances? 

For example when we have the component details and workspace wrapped in a reactflowprovider it means each workflow will have a provider to get its own rfInstance? otherwise this line `const { reactFlowInstance } = useContext(rfContext);` in `[component_details]` (https://github.com/opensearch-project/opensearch-ai-flow-dashboards/pull/57/files#diff-8c84fba8c626d0c26f252d1c729485bf1401aa7ac15ba8015fe942d02ff32c2fR37) wont know which reactflow instance to get right?

Additioonally you have the EuiResizableContainer as a parent of ReactFlowProvider. why not the other way around?

ohltyler: Both `ReactFlowContextProvider` and `ReactFlowProvider` are only intended for a single reactflow instance. We reactflow instance is only set once in `Workspace` via `onInit()` fn provided by reactflow.

To clarify the 2 different uses:
- `ReactFlowContextProvider` is for accessing the instance anywhere within the plugin - see details in #42 
- `ReactFlowProvider` is for accessing the custom hooks provided by reactflow to listen for certain actions taken within the workspace - see this PR description for details

ohltyler: The reason for putting the ReactFlowProvider where I did was just to try to scope it as small as possible to prevent weird side effects.

public/pages/workflow_detail/workspace/component_details.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * Hook provided by reactflow to listen on when nodes are selected / de-selected.
+   * - populate panel content appropriately
+   * - open the panel if a node is selected and the panel is closed
+   */
+  useOnSelectionChange({
+    onChange: ({ nodes, edges }) => {
+      if (nodes && nodes.length > 0) {
+        setSelectedComponent(nodes[0]);
```
amitgalitz, line 51: why is this hard coded to `nodes[0]`?

public/pages/workflow_detail/workspace/resizable_workspace.tsx
amitgalitz: got it so the only place where we have knowledge of all the workflows together is through redux state right?

public/pages/workflow_detail/workspace/component_details.tsx
ohltyler: It is only possible to select one node at once as far as I'm aware - there are certain dynamic grouping options available with reactflow pro which may mean the array would be >1 but I'm unable to find any examples from the docs or any way to trigger >1 node selected when testing things. It may be for future-proofing if added functionality in the free or Pro versions open up multiple selections.

public/pages/workflow_detail/workspace/resizable_workspace.tsx
ohltyler: Yes - workflows are just the standalone backend resources, and we only persist one reactflow instance when a particular workflow is loaded into the details page.

public/pages/workflows/new_workflow/new_workflow.tsx
```diff
@@ -10,6 +10,13 @@ import { UseCase } from './use_case';
 
 interface NewWorkflowProps {}
 
+/**
```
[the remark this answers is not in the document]
ohltyler: Yes to the former. We will probably have a 'create custom workflow' card or something to click on which will load an empty workspace.

public/pages/workflow_detail/workspace_component/workspace_component.tsx
```diff
@@ -23,32 +21,17 @@ interface WorkspaceComponentProps {
...
-        </EuiFlexItem> */}
         {component.inputs?.map((input, index) => {
           return (
             <EuiFlexItem key={index}>
               <InputHandle input={input} data={component} />
             </EuiFlexItem>
           );
         })}
-        <InputFieldList inputFields={fieldsToDisplay} />
+        {/* TODO: finalize from UX what we show in the component itself. Readonly fields? Configure in the component JSON definition? */}
```
amitgalitz, line 34: This bring a good point, we might need to make a decision for each field for each component as some might need drop down, some might need a number and etc. might be good to add on the interface definition

public/pages/workflow_detail/workspace/component_details.tsx
amitgalitz: Ah I see, didn't know there was a pro version. Seemed weird if its an array that we can't grab multiple and then would need some sort of filter function to only set one selected component. maybe add a small comment?

public/pages/workflow_detail/workspace_component/workspace_component.tsx
ohltyler: So this is actually already initially done as part of #32 where we dynamically render based on the input type. That logic is moved here into the side panel.

This comment is about what we actually show in the component itself in the reactflow workspace vs. the rendered required inputs that are in the side panel.

ohltyler: We will not have inputs in the component itself per the latest UX mockups.

amitgalitz: I see, in the video you have above after clicking on text embedding processor, then clicking on `input field` will it open a text box or json?

amitgalitz (approved):

public/pages/workflow_detail/workspace_component/workspace_component.tsx
ohltyler: All 3 of those are text fields - see https://github.com/opensearch-project/opensearch-ai-flow-dashboards/blob/main/public/component_types/processors/text_embedding_processor.ts#L46

Maybe the confusion is that `input_field` is an input field for this example :)

ohltyler: It is using https://oui.opensearch.org/1.1/#/forms/form-controls#text-field

public/pages/workflow_detail/workspace/component_details.tsx
ohltyler: Will add

amitgalitz (approved):

PR #61, opened 2023-10-18
Persist form state and validation in Workspace

ohltyler:
### Description

At a high level, this PR adds 2 main functionalities:
1.  Persist and validate form data/inputs to the components in the Workspace. All of the form inputs and schema are dynamically generated based on what components are in the workspace, and what fields and field types they have specified in their json config. This way, any changes to existing or new components will not require adding any special logic for persisting input values or validating it.
2. Persist selection state in a customizable manner. We keep the node selected when clicking outside of the workspace so users know what data they are editing in the side panel.

More details:
- adds apache-2.0-licensed [formik](https://formik.org/) and MIT-licensed [yup](https://www.npmjs.com/package/yup) packages for form management and validation, respectively. Formik is a widely used & popular library used for form management with react apps, and is already used in other OSD plugins. Formik validation has very good integration with Yup, a library used for defining validation schemas over form data.
- integrates the `TextField` and `SelectField` input components with formik using formik's [Field](https://formik.org/docs/api/field) component
- configures and wraps the entire workspace with the [Formik](https://formik.org/docs/api/formik) component so form data can be accessed in the workspace + details side panel
- adds several util fns to convert component data to and from Formik values and Yup schemas
- refactors and moves `input_fields` into the `component_details` module
- adds a `name` and `value` field to `IComponentField` to generate formik values from component data, or vice versa.
- adds a custom initialization hook in resizable_workspace to dynamically generate the form values and schema based on the workflow's component data
- adds an `onNodesChange()` fn in resizable_workspace used for updating the forms and schema when a component is added or removed from the Workspace. This is done by passing this fn to the workspace, where we use formik's `useStore` custom hook to listen on node count changes, which we then trigger the `onNodesChange` callback fn
- adds a testable 'submit' button on the bottom to test form validation. leaving in for now for the sake of testing submission since there is no finalized UX around this part yet.
- persist selection state by 1/ moving the `ReactFlowProvider` to a higher-level component so we can move `useOnSelectionChange` to resizable_workspace. This simplifies the downstream components. And, 2/ Updates css styling to control when a component is selected or not.

Demo video showing form state:

[screen-capture (19).webm](https://github.com/opensearch-project/opensearch-ai-flow-dashboards/assets/62119629/84a16af3-f612-4dca-930d-ccfd6907b588)

Demo video showing selection state (added after prev. demo video):

[screen-capture (20).webm](https://github.com/opensearch-project/opensearch-ai-flow-dashboards/assets/62119629/e83515dc-09ff-4c46-9ee8-ea92a2222f4f)

### Issues Resolved

Makes progress on #8, #10

### Check List


public/component_types/indices/knn_index.ts
```diff
@@ -73,6 +75,7 @@ export class KnnIndex extends BaseComponent implements IComponent {
       // simple form inputs vs. complex JSON editor
```
amitgalitz, line 75: Have you seen a way to dynamically create a form based on a list of potential fields? Simple way would be key values that you can just keep adding for mapping but maybe something more complicated for nested fields?

ohltyler: That's what this PR achieves. See the helper fns in [resizable_workspace](https://github.com/ohltyler/opensearch-ai-flow-dashboards-1/blob/form/public/pages/workflow_detail/workspace/resizable_workspace.tsx#L87-L124). Lmk if you're referring to something else.

public/pages/workflow_detail/workspace/resizable_workspace.tsx
```diff
@@ -21,48 +32,148 @@ const COMPONENT_DETAILS_PANEL_ID = 'component_details_panel_id';
...
+      // TODO: implement for when a node is added
+    } else if (updatedComponentIds.length < existingComponentIds.length) {
+      existingComponentIds.forEach((existingId) => {
+        if (updatedComponentIds.includes(existingId)) {
+          updatedSchemaObj[existingId] = formSchema.fields[
+            `${existingId}`
+          ] as yup.ObjectSchema<any, any, any>;
+        } else {
+          delete formValues[`${existingId}`];
+        }
```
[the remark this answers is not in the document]
ohltyler: For readability I can try to create some more temp variables, like deep copying the form values. I didn't do that initially since it wasn't necessary, and prevents having to make an extra `setFormValues()` call. Lmk what you think.

public/component_types/indices/knn_index.ts
amitgalitz: I assume for most cases the keys will be defined and I see how we read those from the fields defined for the components. I meant also for cases where we might have users add additional parameters that aren't defined or is that not the plan right now?

public/component_types/indices/knn_index.ts
```diff
@@ -55,6 +55,7 @@ export class KnnIndex extends BaseComponent implements IComponent {
     this.fields = [
       {
         label: 'Index Name',
+        name: 'indexName',
```
amitgalitz, line 58: Even though this is for formik purpose mostly, will this get transfered to the `type` field in the use case or still undecided?

ohltyler: Likely the component [type](https://github.com/opensearch-project/opensearch-ai-flow-dashboards/pull/61/files/c77d128f01685de3a2e4e62f292711c69075a0d2#diff-0d6242f3e63e8da8fa40e9a37de958f05f18c0c5accbedda53221f24c47c5d52L35) will be more of a 1:1 relationship with a use case template step. Since we don't have defined backend step types or component building blocks yet there is nothing concrete.

public/component_types/interfaces.ts
```diff
@@ -35,6 +44,8 @@ export interface IComponentInput {
 export interface IComponentField {
   label: string;
   type: FieldType;
+  name: string;
+  value?: FieldValue;
```
amitgalitz, line 48: Just wondering if we want to have initial values in some cases, lets say we want to make a custom cohere components of sort, will this be the field that gets autopopulated by the formik initial values?

public/component_types/indices/knn_index.ts
amitgalitz: got it, I think I confused myself cause I saw the type here only for the `field` itself which was `select`. Should have expanded to see it was different then the component type you linked too.

public/pages/workflow_detail/component_details/input_fields/select_field.tsx
```diff
@@ -0,0 +1,74 @@
...
+import {
+  IComponentField,
+  WorkspaceFormValues,
+  getInitialValue,
+  isFieldInvalid,
+} from '../../../../../common';
+
+// TODO: Should be fetched from global state.
+// Need to have a way to determine where to fetch this dynamic data.
+const existingIndices = [
```
amitgalitz, line 23: I know this will change to some fetch for all indices as you mention in comment but the select field shouldn't be specific to indices right? Just "existing options"? We can have a list of model IDs here too

public/pages/workflow_detail/workspace/resizable_workspace.tsx
```diff
@@ -21,48 +33,149 @@ const COMPONENT_DETAILS_PANEL_ID = 'component_details_panel_id';
  * panels - the ReactFlow workspace panel and the selected component details panel.
  */
 export function ResizableWorkspace(props: ResizableWorkspaceProps) {
+  // Component details side panel state
   const [isOpen, setIsOpen] = useState<boolean>(true);
```
amitgalitz, line 37: nit: could just change to more descriptive name here like `isDetailsPanelOpen`

public/component_types/interfaces.ts
ohltyler: Yes, right now I populate with initial values based on the field type - see `getInitialValue()`. Rudimentary for now but will expand as add more types or have some more intuitive defaults.

ohltyler: Perhaps some component types may have some defined initial values too. Not adding that for now until that's a necessity.

public/pages/workflow_detail/component_details/input_fields/select_field.tsx
ohltyler: Correct, just have hardcoded values for testing functionality. Has nothing to do with indices.

public/pages/workflow_detail/workspace/resizable_workspace.tsx
ohltyler: Agreed - updated.

public/pages/workflow_detail/workspace/workspace.tsx
```diff
@@ -39,9 +46,17 @@ export function Workspace(props: WorkspaceProps) {
   const reactFlowWrapper = useRef(null);
   const { reactFlowInstance, setReactFlowInstance } = useContext(rfContext);
 
-  const [nodes, setNodes, onNodesChange] = useNodesState([]);
+  const [nodes, setNodes, onNodesChange] = useNodesState<IComponentData>([]);
   const [edges, setEdges, onEdgesChange] = useEdgesState([]);
 
+  // Listener for node additions or deletions to propagate to parent component
+  const nodesLength = useStore(
+    (state) => Array.from(state.nodeInternals.values()).length || 0
```
amitgalitz, line 54: would the nodes.length not be updated to use here?

ohltyler: It's a custom hook that will be updated when the node internals change (node added/deleted/etc.). See docs and similar example: https://reactflow.dev/docs/api/hooks/use-store/

ohltyler: And similar to the example, we listen for when the value changes on our useEffect on line 56-58

public/pages/workflow_detail/workspace/resizable_workspace.tsx
```diff
@@ -21,48 +33,149 @@ const COMPONENT_DETAILS_PANEL_ID = 'component_details_panel_id';
...
+  }, [props.workflow]);
+
+  // Update the form values and validation schema when a node is added
+  // or removed from the workspace.
+  // For the schema, we do a deep clone of the underlying object, and later re-create the schema.
+  // For the form values, we update directly to prevent the form from being reinitialized.
+  function onNodesChange(nodes: ReactFlowComponent[]): void {
+    const updatedComponentIds = nodes.map((node) => node.id);
+    const existingComponentIds = Object.keys(formValues);
+    const updatedSchemaObj = cloneDeep(formSchema.fields) as WorkspaceSchemaObj;
```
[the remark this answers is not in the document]
ohltyler: We extract the fields which is the underlying object, update it if necessary (like removing the component ID on line 119), and re-generate the schema on line 128.

amitgalitz (approved):

PR #135, opened 2024-04-16
Persist created resources; clean up unused tabs

ohltyler:
### Description

This PR updates some of the placeholder tabs to the latest proposed UX changes of smaller scope. It also adds a placeholder 'Resources' tab to display what resources have been created by each workflow for transparency. Note where this list will live exactly is TBD; for now we have in a standalone tab. It can easily be moved and only takes in the single `workflow` prop.

More specific list of changes:
- removes `Overview` page
- removes `Prototype` tab
- removes `Launches` tab (leaving placeholder code as this will be picked back up in the future when workflow versioning is available)
- adds `Resources` tab
- adds `WorkflowResource` basic interface
- adds logic to parse and persist the resources on server-side and within redux store

Demo video:
- creates a workflow (no provisioning) -> no resources created
- provisioning a workflow -> resources created

[screen-capture (28).webm](https://github.com/opensearch-project/dashboards-flow-framework/assets/62119629/89a34b1d-dc1c-4393-90ec-85f54ebd16fa)

### Check List


server/routes/helpers.ts
```diff
@@ -89,10 +96,32 @@ export function getModelsFromResponses(modelHits: any[]): ModelDict {
...
+export function getResourcesCreatedFromResponse(
+  resourcesCreated: any[] | undefined
+): WorkflowResource[] {
+  const finalResources = [] as WorkflowResource[];
+  if (resourcesCreated) {
+    resourcesCreated.forEach((backendResource) => {
+      finalResources.push({
+        id: backendResource.resource_id,
+        type:
+          // @ts-ignore
```
jackiehanyang, line 118: what's the motivation of adding `@ts-igore` here? does it mean it's possible that `WORKFLOW_RESOURCE_TYPE` cannot handle all cases?

ohltyler: See the added enum in `common/interfaces`. The ignore here is because of the type safety issue when referencing an enum with a `string` value instead of `typeof WORKFLOW_RESOURCE_TYPE`. Note I do the same on line 104-105.

When the lookup fails this will propagate that error through `flow_framework_routes_service` correctly and make it visible. We don't want to hide a critical parsing error where the backend and frontend schema become out of sync, so we handle it this way.

jackiehanyang (approved):

Issue #73, opened 2024-02-20
[Workflow editor] Add basic placeholder building blocks


ohltyler:
Develop interfaces and implement some of the basic building blocks with proposed inputs/outputs/fields for support of the initial [neural search use cases](https://opensearch.org/docs/latest/search-plugins/neural-search/)

ohltyler:
Update 4/2: tuned the issue to specifically be initial building blocks for the first use cases - semantic/hybrid/multi-modal search, etc.

ohltyler:
Scoped to different variations of the same ML inference processor block to be applied to initial use cases. Marking as complete.

PR #137, opened 2024-04-18
Add option to select and provision pretrained text embedding models

ohltyler:
### Description

This PR adds the ability to select existing models, or select from available pretrained models, in the text embedding processor component. Specifically:
- adds and updates several ML commons interfaces and constants
- enhances search model logic on server-side to parse out more details from the API responses
- adds some pretrained model constants to be optionally used when building flows
- adds a `ModelField` field type to process all of the logic around model selection, including only showing valid deployed models, and available pretrained models. _Note: the final props and how customizable this field will need to be is TBD. For example, we can customize it to be tailored towards text embedding models vs. other types of models in the future_
- updates form state and form validation to persist more model info instead of just model ID (now model ID + model type (pretrained vs. existing/deployed) + open door for more)
- updates `toTemplateNodes` to dynamically handle the template node creation based on a pretrained vs. existing/deployed model. If a pretrained model is selected, we add a previous step to register and deploy that model, and use the produced model ID in the downstream `create_ingest_pipeline step`. Else, use the existing logic and just use the user/form-supplied model ID directly in a single `create_ingest_pipeline` step.

Misc:
- cleans up routing and URL logic when navigating to the plugin from the global sidebar, in addition to the default rerouting to the workflows page
- removes hyperlink in the base `Flow Framework` breadcrumb
- changes the provisioned workflow state slightly by removing the readonly guardrails and allowing the components to be selectable to see the details. Adds a warning callout when values are changed but the workflow is provisioned, and still enforces the workflow to be deprovisioned before changes can be saved. Adds TODOs if we want to revert this change
- enforces uniqueness on resource ID list to prevent duplicates which can happen on backend in certain edge cases (comments in code about it)

Demo video:
- creates workflow with pretrained model
- provisions workflow to create the model and use it in the downstream creation of an ingest pipeline
- views the workflow resources, including the newly created model

[screen-capture (29).webm](https://github.com/opensearch-project/dashboards-flow-framework/assets/62119629/23a4e12d-04b8-42a5-9f9e-efd3140d9023)

### Issues Resolved
Makes progress on #73 

### Check List

### DIFF
--- a/common/constants.ts
+++ b/common/constants.ts
@@ -4,6 +4,7 @@
  */
 
 import {
+  MODEL_ALGORITHM,
   PRETRAINED_MODEL_FORMAT,
   PretrainedSentenceTransformer,
   WORKFLOW_STATE,
@@ -68,6 +69,7 @@ export const ROBERTA_SENTENCE_TRANSFORMER = {
   shortenedName: 'all-distilroberta-v1',
   description: 'A sentence transformer from Hugging Face',
   format: PRETRAINED_MODEL_FORMAT.TORCH_SCRIPT,
+  algorithm: MODEL_ALGORITHM.TEXT_EMBEDDING,
   version: '1.0.1',
   vectorDimensions: 768,
 } as PretrainedSentenceTransformer;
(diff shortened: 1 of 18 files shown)


amitgalitz:
Might have missed it in the code but also if its a pretrained model it has the info like dimensions in the model metadata for further steps that need that

amitgalitz (approved):

ohltyler:
> Might have missed it in the code but also if its a pretrained model it has the info like dimensions in the model metadata for further steps that need that

Right. I have this in line 196 in `common/interfaces`. Now that we have it persisted, we can utilize it dynamically during template construction later on. I still have it hardcoded atm, but there's a TODO tracker for it

Issue #23, opened 2023-09-05
[META] Workflow editor


ohltyler:
Introduce a workflow editor containing a drag-and-drop workspace where users can stitch together generic building blocks to build out their applications. When validated, they can launch their workflow to create any necessary resources and begin to prototype.

Proposed mock as of 2/19 from @opensearch-project/opensearch-ux:

![Screenshot 2024-02-19 at 4 38 19 PM](https://github.com/opensearch-project/dashboards-flow-framework/assets/62119629/dd1d842f-4f3e-442e-9e38-63bd42119abd)

- [x] #10 
- [x] #20 
- [x] #5 
- [x] #73 
- [x] #74 
- [x] #75 
- [x] #82

ohltyler:
Basic components and functionality complete. Drag-and-drop is deprioritized so removed that from this meta issue. Resolving.

PR #188, opened 2024-06-20
Integrate query into form (not persisted in config)

ohltyler:
### Description

Similar to the way we persist the ingest documents, we persist the query in the same form. This means:
1/ integrating with the overall form for schema/validation and temporary persistence
2/ not integrating with the underlying, persisted, indexed, config.

More details:
- persist the query state at the top-level `ResizableWorkspace` like the ingest docs
- update the query request form `ConfigureSearchRequest` to be similar to the ingest docs form, utilizing the `JsonField` component (which handles the auto-formatting as well)
- update `validateAndRunQuery()` to ensure there is a valid, non-empty query, when attempting to click 'Run query' button
- final updates to the formik and schema conversion fns to include the query field

Demo video showing validation, auto-formatting, persistence across page navigation, and finally the query execution:

[screen-capture (39).webm](https://github.com/opensearch-project/dashboards-flow-framework/assets/62119629/19d68be0-e832-4b5a-8e12-f521495285b3)

### Issues Resolved
Makes progress on #23 

### Check List

### DIFF
--- a/public/utils/config_to_schema_utils.ts
+++ b/public/utils/config_to_schema_utils.ts
@@ -39,6 +39,30 @@ function ingestConfigToSchema(
   return yup.object(ingestSchemaObj);
 }
 
+function indexConfigToSchema(indexConfig: IndexConfig): Schema {
+  const indexSchemaObj = {} as { [key: string]: Schema };
+  indexSchemaObj['name'] = getFieldSchema(indexConfig.name.type);
+  indexSchemaObj['mappings'] = getFieldSchema(indexConfig.mappings.type);
+  indexSchemaObj['settings'] = getFieldSchema(indexConfig.settings.type);
+  return yup.object(indexSchemaObj);
+}
+
+function searchConfigToSchema(
+  searchConfig: SearchConfig | undefined
+): ObjectSchema<any> {
+  const searchSchemaObj = {} as { [key: string]: Schema };
+  if (searchConfig) {
+    searchSchemaObj['request'] = getFieldSchema('json');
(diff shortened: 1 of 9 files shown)


public/utils/config_to_schema_utils.ts
```diff
@@ -60,15 +47,35 @@ function indexConfigToSchema(indexConfig: IndexConfig): Schema {
...
+  const processorsSchemaObj = {} as { [key: string]: Schema };
+  processorsConfig.processors.forEach((processorConfig) => {
+    const processorSchemaObj = {} as { [key: string]: Schema };
+    processorConfig.fields.forEach((field) => {
+      processorSchemaObj[field.id] = getFieldSchema(field.type);
+    });
+    processorsSchemaObj[processorConfig.id] = yup.object(processorSchemaObj);
+  });
+
+  return yup.object(processorsSchemaObj);
```
jackiehanyang, line 76: I could be wrong as I'm not very familiar with the Yup library, had to look up the Yup library documentation about `yup.object` to understand what we're trying to do here - define the schema. I see there's another method - `yup.object().shape`, that defines the structure and validation rules for an object. Should we change to use `yup.object().shape` here as `yup.object()` just defines the structure without any constraints?

ohltyler: The parameters passed in `object` are where the constraints are set - see [documentation](https://github.com/jquense/yup?tab=readme-ov-file#object). Passing verbosely with `shape` achieves the same thing, but prefer the short-form to keep it simpler.

jackiehanyang (approved):

PR #229, opened 2024-07-22
Integrate with JSONPath; complete input transform (ingest)

ohltyler:
### Description

This PR completes the end-to-end flow of input transforms on the ingest side, by way of adding JSONPath parsing support within the ML inference input map configs. More specifically:
- adds MIT-licensed JSONPath (base [JS](https://www.npmjs.com/package/jsonpath) and [TS](https://www.npmjs.com/search?q=%40types%2Fjsonpath)) packages to the plugin
- finishes base functionality of `InputTransformModal` - adds the map form field within the modal, completes generation logic of the output JSON after applying JSONPath and/or dot notation to the specified inputs in the map. Note the logic used here is the same as the ML inference processor logic, when parsing the map config. If no `$.` JSONPath base object selector is specified, it defaults to using dot notation. We do the same logic to produce consistent results when actually used in the ML inference processor
- removes cancel/save of `InputTransformModal` to just be `close`, since the autosave will handle it
- adds simple/advanced toggle in the ML processor form inputs component, refactors some of the buttons and map field inputs based on simple vs. advanced being selected
- adds `keyPlaceholder`/`valuePlaceholder` props to `MapField` to be able to reuse in the output maps and for other contexts (ingest / search req / search resp) later on

Testing
- confirmed consistent parsing for dot notation / JSONPath / invalid inputs and combinations of all of them, when input in the transform modal, and when actually executed in the processor with an ingest command

Demo video, showing end-to-end flow of configuring an ML inference processor's input_map using both dot notation and JSONPath, getting sample results in the advanced transform modal, and finally running a concrete ingest command against the configured processor:

[screen-capture (8).webm](https://github.com/user-attachments/assets/252ef4c8-faf4-43a1-a019-86531090096e)

### Issues Resolved

Makes progress on #23 

### Check List

### DIFF
--- a/public/pages/workflow_detail/workflow_inputs/processor_inputs/input_transform_modal.tsx
+++ b/public/pages/workflow_detail/workflow_inputs/processor_inputs/input_transform_modal.tsx
@@ -5,12 +5,11 @@
 
 import React, { useState } from 'react';
 import { useFormikContext, getIn } from 'formik';
-import { isEmpty } from 'lodash';
+import { isEmpty, get } from 'lodash';
 import jsonpath from 'jsonpath';
 import {
   EuiButton,
   EuiButtonEmpty,
-  EuiCallOut,
   EuiCodeBlock,
   EuiFlexGroup,
   EuiFlexItem,
@@ -26,6 +25,7 @@ import {
   IConfigField,
   IProcessorConfig,
   IngestPipelineConfig,
+  JSONPATH_ROOT_SELECTOR,
   PROCESSOR_CONTEXT,
   SimulateIngestPipelineDoc,
   SimulateIngestPipelineResponse,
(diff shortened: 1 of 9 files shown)


public/pages/workflow_detail/workflow_inputs/processor_inputs/input_transform_modal.tsx
```diff
@@ -116,24 +124,105 @@ export function InputTransformModal(props: InputTransformModalProps) {
...
-                  autoScrollEditorIntoView: true,
-                }}
-                tabSize={2}
+              <MapField
+                field={props.inputMapField}
+                fieldPath={props.inputMapFieldPath}
+                label="Input map"
+                helpText={`An array specifying how to map fields from the ingested document to the model’s input.`}
+                helpLink={
+                  'https://opensearch.org/docs/latest/ingest-pipelines/processors/ml-inference/#configuration-parameters'
```
owaiskazi19, line 139: We should have the links in a common file or the constant file

owaiskazi19 (approved):

public/pages/workflow_detail/workflow_inputs/processor_inputs/input_transform_modal.tsx
ohltyler: Good point, will move in the next PR

PR #244, opened 2024-07-25
Support multiple input/output maps for ML processors

ohltyler:
### Description

This PR expands the form inputs for the input map / output map fields for ML processors, to be a _list_ of maps, instead of a single map. This is so multiple predict/inference calls can happen within a single processor's execution. At a high level, this means changing the form, schema, and components to maintain an arr of maps instead of a single one.

Additionally, In the advanced transform modals, users can now select which predict/inference call they want to test out and see test transformation results of.

More specific details:
- add new `mapArray` config field type
- add new `MapArrayField` which takes in a `Formik`'s arr field and maintains an arr of individual `MapField`s
- update the transform modals to maintain a specific map/prediction to test with. Adds in a selector for users to choose which map/prediction they want
- minor refactoring updates to the conversion fns for handling the new `mapArray` field type, used in the ML processor's `inputMap`/`outputMap` fields

Testing
- confirmed arrays of input_maps and output_maps work as expected in the simulate pipeline calls within the transform modals, as well as the actual ingest calls and checking the indexed data
- confirmed edge cases (some empty/mismatching arr sizes) for both input and output transforms work as expected

Demo video, showing configuration of multiple prediction/inference mappings configured in the input and output maps, validation of mismatching, and transformation testing within the input transform modal and output transform modals:

[screen-capture (12).webm](https://github.com/user-attachments/assets/0e00feef-c2a9-43c0-af75-6eff7c0aa98e)

### Issues Resolved
Makes progress on #23 

### Check List

### DIFF
--- a/public/pages/workflow_detail/workflow_inputs/input_fields/map_array_field.tsx
+++ b/public/pages/workflow_detail/workflow_inputs/input_fields/map_array_field.tsx
@@ -33,6 +33,7 @@ interface MapArrayFieldProps {
   keyPlaceholder?: string;
   valuePlaceholder?: string;
   onFormChange: () => void;
+  onMapAdd?: (curArray: MapArrayFormValue) => void;
   onMapDelete?: (idxToDelete: number) => void;
 }
 
@@ -49,6 +50,9 @@ export function MapArrayField(props: MapArrayFieldProps) {
     setFieldValue(props.fieldPath, [...curMapArray, []]);
     setFieldTouched(props.fieldPath, true);
     props.onFormChange();
+    if (props.onMapAdd) {
+      props.onMapAdd(curMapArray);
+    }
   }
 
   // Deleting a map
--- a/public/pages/workflow_detail/workflow_inputs/input_fields/map_array_field.tsx
+++ b/public/pages/workflow_detail/workflow_inputs/input_fields/map_array_field.tsx
@@ -71,6 +71,7 @@ export function MapArrayField(props: MapArrayFieldProps) {
       {({ field, form }: FieldProps) => {
         return (
           <EuiFormRow
+            fullWidth={true}
             key={props.fieldPath}
             label={props.label}
             labelAppend={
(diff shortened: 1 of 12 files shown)


public/pages/workflow_detail/workflow_inputs/input_fields/map_array_field.tsx
```diff
@@ -0,0 +1,149 @@
...
+  valuePlaceholder?: string;
+  onFormChange: () => void;
+  onMapAdd?: (curArray: MapArrayFormValue) => void;
+  onMapDelete?: (idxToDelete: number) => void;
+}
+
+/**
+ * Input component for configuring an array of field mappings
+ */
+export function MapArrayField(props: MapArrayFieldProps) {
```
jackiehanyang, line 43: nit: maybe consider to wrap it with memo to ensure that the component only re-renders when its props change

public/pages/workflow_detail/workflow_inputs/processor_inputs/input_transform_modal.tsx
```diff
@@ -59,10 +66,26 @@ export function InputTransformModal(props: InputTransformModalProps) {
...
   // source input / transformed output state
   const [sourceInput, setSourceInput] = useState<string>('[]');
-  const [transformedOutput, setTransformedOutput] = useState<string>('[]');
+  const [transformedOutput, setTransformedOutput] = useState<string>('{}');
 
   // get the current input map
-  const map = getIn(values, `ingest.enrich.${props.config.id}.inputMap`);
+  const map = getIn(
+    values,
+    `ingest.enrich.${props.config.id}.inputMap`
```
jackiehanyang, line 74: will the path always exist? should we handle the case where the path may not exist?

public/pages/workflow_detail/workflow_inputs/processor_inputs/input_transform_modal.tsx
```diff
@@ -59,10 +66,26 @@ export function InputTransformModal(props: InputTransformModalProps) {
...
+  const outputOptions = map.map((_, idx) => ({
+    value: idx,
+    text: `Prediction ${idx + 1}`,
+  })) as EuiSelectOption[];
+  const [selectedOutputOption, setSelectedOutputOption] = useState<
+    number | undefined
+  >(
+    get(outputOptions, '0.value') !== undefined
+      ? (outputOptions[0].value as number)
+      : undefined
```
jackiehanyang, line 87: nit - same as a couple of occurrences below: this can be simplified as `outputOptions[0]?.value ?? undefined`

public/pages/workflow_detail/workflow_inputs/input_fields/map_array_field.tsx
ohltyler: Good point! Since this is still in rapid development, I'm going to leave as-is for now, and will consider that a standalone task to audit all components before release. Created a tracking issue and triaged for 2.17: #247

public/pages/workflow_detail/workflow_inputs/processor_inputs/input_transform_modal.tsx
ohltyler: It does always exist. But, I've actually simplified this since the full path is passed in as a prop already. see latest commit.

public/pages/workflow_detail/workflow_inputs/processor_inputs/input_transform_modal.tsx
ohltyler: sure, will update.

jackiehanyang (approved):

PR #253, opened 2024-07-31
Support MDS on workflows and workflow_detail pages

### LINKED ISSUE #228: [FEATURE] Support Multiple Data Source in Flow Framework Dashboards Plugin
**Is your feature request related to a problem?**
With multiple data source enabled, users of OpenSearch Dashboards are able to create data source connections and query local and remote OpenSearch clusters. Enabling multiple data source at dashboards plugin would ensure users that have been using the multiple data source feature to have a consistent experience in the dashboard, and allows users to use the same dashboard for different clusters.

**What solution would you like?**
Support multiple data source in the anomaly detection dashboards plugin and allow user to use one dashboard for different backend clusters

**What alternatives have you considered?**

**Do you have any additional context?**

saimedhi:
### Description

- Added support for using MDS on workflows and workflow_detail pages
- Update all route path to make it accept dataSourceId when it has truthy value

### Issues Resolved

closes #228 

### Check List

  - [ ] All tests pass


saimedhi:
https://github.com/user-attachments/assets/0c5091fe-4aba-4685-8613-1092f4071861

jackiehanyang:
Could you mark the PR as a draft until you have completed the end-to-end testing? It includes:
- Turning on/off these two feature flags to ensure backward compatibility works fine.
```
# Set the value of this setting to true to enable multiple data source feature.
#data_source.enabled: false
# Set the value of this setting to true to hide local cluster in data source feature.
#data_source.hideLocalCluster: false
```
- Switching between two remote data source clusters to ensure workflow-related data and data source IDs are shown correctly (in url, breadcrumbs.. any possible places).
- Proving that the workflow you created when remote data source A was connected can also be found in cluster A.
- Since we don't have tests here, try to click on all possible buttons to test all functionalities. For example, the delete and link button on the Manage Workflows page. 
- Making sure CIs are passing

public/services.ts
```diff
@@ -4,11 +4,32 @@
...
 
 export const [getCore, setCore] = createGetterSetter<CoreStart>('Core');
 
 export const [getRouteService, setRouteService] = createGetterSetter<
   RouteService
 >('');
+
+export const [getSavedObjectsClient, setSavedObjectsClient] =
+  createGetterSetter<CoreStart['savedObjects']['client']>('SavedObjectsClient');
+
```
ohltyler, line 23: dup: we shouldn't need saved objects client

public/services.ts
```diff
@@ -4,11 +4,32 @@
...
+export const [getDataSourceEnabled, setDataSourceEnabled] =
+  createGetterSetter<DataSourceEnabled>('DataSourceEnabled');
+
+export const [getNotifications, setNotifications] =
+  createGetterSetter<NotificationsStart>('Notifications');
+
+export default {
+  getSavedObjectsClient,
+};
```
ohltyler, line 35: can remove this.

public/utils/helpers.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+import { MDSQueryParams } from 'server/types';
```
ohltyler, line 12: dup: make sure to update this when moved into `commons/` somewhere.

public/utils/helpers.ts
ohltyler, line 1: we already have `public/utils/utils.ts` for maintaining these types of reusable fns - can you add there instead, for consistency?

public/utils/helpers.ts
```diff
@@ -0,0 +1,42 @@
...
+  }): MDSQueryParams => {
+    const queryParams = queryString.parse(location.search);
+    const dataSourceId = queryParams.dataSourceId;
+    return { dataSourceId: typeof dataSourceId === 'string' ? dataSourceId : '' };
+  };
+
+export const constructHrefWithDataSourceId = (
+    basePath: string,
+    dataSourceId: string = '',
+    withHash: Boolean
```
ohltyler, line 28: nit: `Boolean` -> `boolean`

server/plugin.ts
```diff
@@ -24,9 +24,16 @@ import {
   registerMLRoutes,
   MLRoutesService,
 } from './routes';
+import { DataSourcePluginSetup } from '../../../src/plugins/data_source/server/types';
+import { DataSourceManagementPlugin } from '../../../src/plugins/data_source_management/public';
 
 import { ILegacyClusterClient } from '../../../src/core/server/';
 
+export interface FFPluginSetupDependencies {
```
ohltyler, line 32: nit: for consistency, let's call this `FlowFrameworkPluginSetupDependencies`

server/plugin.ts
```diff
@@ -57,9 +67,14 @@ export class FlowFrameworkDashboardsPlugin
       }
     );
 
-    const opensearchRoutesService = new OpenSearchRoutesService(client);
-    const flowFrameworkRoutesService = new FlowFrameworkRoutesService(client);
-    const mlRoutesService = new MLRoutesService(client);
+    const dataSourceEnabled = !!dataSource;
+    if (dataSourceEnabled) {
+      dataSource.registerCustomApiSchema(flowFrameworkPlugin);
+      dataSource.registerCustomApiSchema(mlPlugin);
```
ohltyler, line 73: what does this do?

server/routes/flow_framework_routes_service.ts
```diff
@@ -58,6 +60,19 @@ export function registerFlowFrameworkRoutes(
     flowFrameworkRoutesService.getWorkflow
   );
 
+  router.get(
+    {
+      path: `${BASE_NODE_API_PATH}/{dataSourceId}/workflow/{workflow_id}`,
```
ohltyler, line 65: nit: common pattern has the vars in snake case - could you update `dataSourceId` -> `data_source_id`?

server/utils/helpers.ts
```diff
@@ -0,0 +1,56 @@
...
+export function getClientBasedOnDataSource(
+  context: RequestHandlerContext,
+  dataSourceEnabled: boolean,
+  request: OpenSearchDashboardsRequest,
+  dataSourceId: string,
+  client: ILegacyClusterClient
+): (
+  endpoint: string,
+  clientParams?: Record<string, any>,
+  options?: LegacyCallAPIOptions
```
ohltyler, line 31: I don't see this parameter being used in the fn, can it be removed?

ohltyler (commented):
Overall great start, thanks a lot for helping! Generally just some syntax issues and best practices, no big concerns.

public/pages/workflows/workflows.tsx
```diff
@@ -81,22 +109,81 @@ export function Workflows(props: WorkflowsProps) {
...
+                ? undefined
+                : [{ id: MDSOverviewState.selectedDataSourceId }],
+            savedObjects: getSavedObjectsClient(),
+            notifications: getNotifications(),
+            onSelectedDataSources: (dataSources) =>
+              handleDataSourceChange(dataSources),
+          }}
+        />
+      );
+    }, [getSavedObjectsClient, getNotifications(), props.setActionMenu]);
```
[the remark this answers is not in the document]
jackiehanyang: `useMemo()` is introduced here after the MDS team made a change on their side that caused the plugin page, which consumes this component, to constantly refresh. So MDS team provided the guideline that `DataSourceMenu` component needs to be memoized with its configuration, ensuring it only re-renders if one of its dependencies changes. 

`MDSOverviewState` changes are better handled separately in the `useEffect()` hook

public/route_service.ts
```diff
@@ -33,53 +34,59 @@ import {
...
+  },
+  dataSourceId: string|undefined) => Promise<any | HttpFetchError>;
 }
 
 export function configureRoutes(core: CoreStart): RouteService {
   return {
-    getWorkflow: async (workflowId: string) => {
+    getWorkflow: async (workflowId: string, dataSourceId: string|undefined) => {
       try {
+        const url = dataSourceId ? `${BASE_NODE_API_PATH}/${dataSourceId}/workflow` : GET_WORKFLOW_NODE_API_PATH;
```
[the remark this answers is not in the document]
saimedhi: I'll raise a GitHub issue and explore ways to simplify or automate this in the upcoming PRs.

server/plugin.ts
saimedhi: During plugin set up, in order to register the plugin API schema/spec, we do the following dataSource.registerCustomApiSchema(flowFrameworkPlugin)


Reference: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/7578#issue-2438193350

saimedhi:
Not ready for review. Will upload the testing screen recording in some time.

public/plugin.ts
```diff
@@ -45,10 +60,15 @@ export class FlowFrameworkDashboardsPlugin
...
     });
+    setDataSourceManagementPlugin(plugins.dataSourceManagement);
+    const enabled = !!plugins.dataSource;
+    setDataSourceEnabled({ enabled });
     return {};
   }
 
   public start(core: CoreStart): FlowFrameworkDashboardsPluginStart {
+    setNotifications(core.notifications);
+    setSavedObjectsClient(core.savedObjects.client);
```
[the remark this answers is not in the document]
saimedhi: Tyler, savedObjectsClient is needed for MDS support as of my knowledge.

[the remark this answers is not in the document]
ohltyler: Ah! I see it now in the `DataSourceMenu` component, got it.

public/services.ts
ohltyler: ignore this

public/services.ts
ohltyler: ignore this

public/pages/workflows/new_workflow/new_workflow.tsx
```diff
@@ -45,7 +47,7 @@ export function NewWorkflow(props: NewWorkflowProps) {
 
   // initial state
   useEffect(() => {
-    dispatch(getWorkflowPresets());
+    dispatch(getWorkflowPresets({ dataSourceId }));
```
ohltyler, line 50: This call is actually only hitting server-side, and never hits any OpenSearch API, hence does not need any data source ID propagation. It is just to fetch the preset workflows we have stored under `server/resources/templates`. Can you remove here, in the reducer, and on server-side?

public/pages/workflows/workflows.tsx
```diff
@@ -54,6 +76,13 @@ function replaceActiveTab(activeTab: string, props: WorkflowsProps) {
  */
 export function Workflows(props: WorkflowsProps) {
   const dispatch = useAppDispatch();
+  const location = useLocation();
+  const queryParams = getDataSourceFromURL(location);
+  const dataSourceEnabled = getDataSourceEnabled().enabled;
+  const [MDSOverviewState, setMDSOverviewState] = useState<MDSStates>({
+    queryParams,
+    selectedDataSourceId: queryParams.dataSourceId,
+  });
```
ohltyler, line 85: I think this can be simplified a bunch of ways. `MDSStates` is redundantly capturing the data source ID, both as a nested field under `queryParams.dataSourceId` and in the standalone `selectedDataSourceId` field. Only `selectedDataSourceId` is ever used. This could instead just persist the data source ID string:
```
const [dataSourceId, setDataSourceId] = useState<string>(queryParams.dataSourceId);
```

All of the references to it can be changed from `MDSOverviewState.selectedDataSourceId` -> `dataSourceId`.

The one place where this state is updated can just be changed to
```
setDataSourceId(dataSourceId)
```

common/interfaces.ts
```diff
@@ -16,6 +16,14 @@ export type Index = {
...
  ********** WORKFLOW TYPES/INTERFACES **********
  */
+export type MDSQueryParams = {
+  dataSourceId: string;
+};
+
+export interface MDSStates {
+  queryParams: MDSQueryParams;
+  selectedDataSourceId?: string;
+}
```
ohltyler, line 26: I think this can be removed, or at the very least simplified. It is redundantly capturing `dataSourceId` both nested under `queryParams.dataSourceId` and in `selectedDataSourceId`.

public/pages/workflows/workflows.tsx
ohltyler: And I think that means `MDSStates` can be removed altogether; I don't think it's used anywhere else but you can double check.

public/pages/workflows/workflows.tsx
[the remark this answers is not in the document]
ohltyler: I'm still confused on the values it is listening on. Shouldn't it just update when the dataSourceId is updated? Why listen on changes to `getSavedObjectsClient, getNotifications(), props.setActionMenu`? @jackiehanyang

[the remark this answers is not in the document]
ohltyler: It seems it could be simplified to just `[dataSourceId]`

public/route_service.ts
```diff
@@ -33,80 +34,124 @@ import {
...
+  catIndices: (
+    pattern: string,
+    dataSourceId?: string
+  ) => Promise<any | HttpFetchError>;
   searchIndex: (
     index: string,
     body: {},
+    dataSourceId?: string,
     searchPipeline?: string
   ) => Promise<any | HttpFetchError>;
```
ohltyler, line 82: Having multiple optional fields that are not defined under some interface can lead to issues. For example, calling searchIndex with a valid search pipeline, but no data source ID - e.g., `searchIndex('abc', {query: match_all: {}}, 'my-search-pipeline'` would treat `my-search-pipeline` as the `dataSourceId` param and fail.

For these cases, let's wrap all of the params in an interfaced obj - something like:
```
 searchIndex: (
{
    index: string,
    body: {},
    dataSourceId?: string,
    searchPipeline?: string
  } )
```

public/route_service.ts
```diff
@@ -33,80 +34,124 @@ import {
...
+  ingest: (
+    index: string,
+    doc: {},
+    dataSourceId?: string
+  ) => Promise<any | HttpFetchError>;
+  bulk: (
+    body: {},
+    dataSourceId?: string,
+    ingestPipeline?: string
+  ) => Promise<any | HttpFetchError>;
```
ohltyler, line 92: same as above comment on `searchIndex`.

public/services.ts
```diff
@@ -4,11 +4,40 @@
...
+  DataSourceEnabled
+>('DataSourceEnabled');
+
+export const [getNotifications, setNotifications] = createGetterSetter<
+  NotificationsStart
+>('Notifications');
+
+export default {
+  getSavedObjectsClient,
+};
```
ohltyler, line 43: this can be removed.

public/plugin.ts
```diff
@@ -9,21 +9,35 @@ import {
...
+  setNotifications,
+} from './services';
 import { configureRoutes } from './route_service';
+import { DataSourceManagementPluginSetup } from '../../../src/plugins/data_source_management/public';
+import { DataSourcePluginSetup } from '../../../src/plugins/data_source/public';
+
+export interface FlowFrameworkDashboardsSetupDeps {
+  dataSourceManagement: DataSourceManagementPluginSetup;
+  dataSource: DataSourcePluginSetup;
+}
```
ohltyler, line 29: Can this be moved back into `public/types.ts` and keep the original name of `FlowFrameworkDashboardsPluginSetup` for consistency?

public/utils/constants.ts
```diff
@@ -18,3 +18,11 @@ export const BREADCRUMBS = Object.freeze({
...
   WORKFLOWS: { text: 'Workflows', href: `#${APP_PATH.WORKFLOWS}` },
 });
+
+export const MDS_BREADCRUMBS = Object.freeze({
+  FLOW_FRAMEWORK: { text: 'Flow Framework' },
+  WORKFLOWS: (dataSourceId?: string) => ({
+    text: 'Workflows',
+    href: `#${APP_PATH.WORKFLOWS}?dataSourceId=${dataSourceId}`,
+  }),
+});
```
ohltyler, line 28: I don't think we need `MDS_BREADCRUMBS` at all. We should just be able to dynamically handle the `dataSourceId` directly in the existing breadcrumbs and not have duplicate code. Something like:
```
export const BREADCRUMBS = Object.freeze({
  FLOW_FRAMEWORK: { text: 'Flow Framework' },
  WORKFLOWS: (dataSourceId?: string) => ({
    text: 'Workflows',
    href: constructHrefWithDataSourceId(APP_PATH.WORKFLOWS, dataSourceId, true),
  }),
});
```

public/utils/utils.ts
```diff
@@ -228,3 +231,36 @@ export function parseModelOutputs(
...
+  if (dataSourceEnabled && dataSourceId !== undefined) {
+    url.set('dataSourceId', dataSourceId);
+  }
+  // we share this helper function to construct the href with dataSourceId
+  // some places we need to return the url with hash, some places we don't need to
+  // so adding this flag to indicate if we want to return the url with hash
+  if (withHash) {
+    return `#${basePath}?${url.toString()}`;
+  }
+  return `${basePath}?${url.toString()}`;
```
ohltyler, line 259: can be simplified to 
```
return `${withHash ? '#' : ''}${basePath}?${url.toString()}`;
```

server/utils/helpers.ts
```diff
@@ -0,0 +1,32 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
```
ohltyler, line 10: can use the shorthand license header:
```
/*
 * Copyright OpenSearch Contributors
 * SPDX-License-Identifier: Apache-2.0
 */
```

jackiehanyang:
Thanks for the change! Two things I noticed from the screen recording:
1. I don't see `dataSourceId=` in the url on `new_workflow` page when local cluster is selected. Let's add that to keep all pages url in consistent when the feature flag is enabled
2. Could you click on the delete and link icon button on the `Workflows` page and to see if they all work as expected?

ohltyler:
@saimedhi looking great! Once my remaining comments on `public/route_service.ts` are addressed and you can confirm all of the functionality is still working as expected, I am happy to approve :)

saimedhi:
[screen-capture (2).webm](https://github.com/user-attachments/assets/10e811fb-fda4-4a32-a159-4c323a9fd826)

public/pages/workflows/workflows.tsx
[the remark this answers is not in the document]
saimedhi: @ohltyler, 
By testing the MDS Picker, I can see that the following are required to set the dataSourceId properly in url, when selecting a data source present in the MDS Picker:
`getSavedObjectsClient, getNotifications(), props.setActionMenu`

ohltyler:
Changes LGTM! One more request to add details in the PR description, including some of the low-level implementation details. See existing merged PRs as an example. This makes it much easier to track down issues that may come up in the future, and keeps a detailed log of what and when changes were made.

jackiehanyang (changes requested):
Screenshot 2024-08-07 at 12 40 57
Screenshot 2024-08-07 at 12 41 06

Remote data source should be consistent in breadcrumbs. When go back to workflow page, `local cluster` should be selected instead of selecting the default data source. Could you address this bug?

saimedhi:
> Remote data source should be consistent in breadcrumbs. When go back to workflow page, `local cluster` should be selected instead of selecting the default data source. Could you address this bug?

@jackiehanyang, I will fix it in a followup PR. Thank you.  

- I will make sure when we go back to workflow page, data source selected will be retained .

public/app.tsx
```diff
@@ -61,12 +75,22 @@ export const FlowFrameworkDashboardsApp = (props: Props) => {
             path={APP_PATH.WORKFLOW_DETAIL}
             render={(
               routeProps: RouteComponentProps<WorkflowDetailRouterProps>
-            ) => <WorkflowDetail {...routeProps} />}
+            ) => (
+              <WorkflowDetail
+                setActionMenu={setHeaderActionMenu}
+                landingDataSourceId={dataSourceId}
```
jackiehanyang, line 81: you don't need this props, and I don't see any usage of this props later. I have this for AD because AD has multiple pages. So I need this prop for AD to track the dataSourceId for landing pages to differentiate:
- if the page is opened as a landing page, we select the default remote data source
- if the page is opened by being redirected from another page, we keep the dataSourceId consistent as the previous page instead of selecting the default remote data source. 

None of this is applicable to Flow Framework, so please remove unnecessary and unused props.

public/utils/utils.ts
```diff
@@ -229,6 +232,36 @@ export function parseModelOutputs(
...
+}): MDSQueryParams => {
+  const queryParams = queryString.parse(location.search);
+  const dataSourceId = queryParams.dataSourceId;
+  return { dataSourceId: typeof dataSourceId === 'string' ? dataSourceId : '' };
+};
+
+export const constructHrefWithDataSourceId = (
+  basePath: string,
+  dataSourceId: string = '',
+  withHash: boolean
```
jackiehanyang, line 246: you don't need this flag. I see you always pass true here

saimedhi:
> > Remote data source should be consistent in breadcrumbs. When go back to workflow page, `local cluster` should be selected instead of selecting the default data source. Could you address this bug?
> 
> @jackiehanyang, I will fix it in a followup PR. Thank you.
> 
> * I will make sure when we go back to workflow page, data source selected will be retained .

[screen-capture (3).webm](https://github.com/user-attachments/assets/b482f6da-84e2-40aa-8644-4706a3468997)

jackiehanyang:
I see a bug - when you opening flow framework from the side nav, default remote data source should be selected. 

Screenshot 2024-08-08 at 11 34 59

Please thoroughly test the following scenario:
- when you opening flow framework from the side nav, default remote data source should be selected
- when you opening flow framework from a bookmarked url, the data source that's in the url should be selected
- when you select a non default remote data source, it should still be selected if you refresh the page, instead of jumping back to the default one.
- data source should be consistent through breadcrumbs
- data source should be consistent through in-app side nav
- considering the project goals and customer perspective, test all possible workflows to ensure that multiple data sources are supported correctly.

PR #268, opened 2024-08-06
Support fine-grained provisioning; handle more edge cases on search flow

### LINKED ISSUE #267: Integrate with the fine-grained provisioning API
Integrate with the fine-grained provisioning / reprovision API to smoothly update resources in-place in the search flow. This will allow adding a search pipeline resource to the workflow without deprovisioning existing ingest data and losing existing documents.

ohltyler:
### Description

This PR improves the search flow by allowing fine-grained provisioning (creating a search pipeline and attaching to an existing index) instead of the limitation from before, forcing full index deletion and re-creation. Now data is persisted and can be searched on in the search flow end-to-end. Additionally, cleans up some edge cases and improves the search flow experience overall, making it fully functional end-to-end now.

More details:
- adds `reprovision` parameter on the update workflow API call to support fine-grained provisioning (ref: https://github.com/opensearch-project/flow-framework/pull/804)
- persists several other fields in the UI config, including 1/ ingest pipeline name, 2/ search pipeline name, and 3/ search index name. The pipeline names are created initially & persisted so we have consistent template node IDs during subsequent updates (before, we were re-generating every time, in which case node IDs did not match, which caused fine-grained provisioning to fail, as it perceived this as node deletions, since the node IDs from the existing template were no longer present). The search index name is a way to persist a user-chosen index (if they skip ingest) and persist it in the UI. This may be expanded upon in the future if we want to pull existing UI configs into an augmented template for users as an option.
- blocks 'Run query' if there is no search pipeline config (the request and response processor configs are empty). This prevents several different edge cases and confusing user experiences otherwise
- adds fine-grained validation: only validate the relevant form fields depending on users being in the context of ingest or search.
- removed bunch of TODO comments :)

Testing:
Tested following scenarios:
- create ingest resources, add on search resources
- skip ingest resources, add on search resources
- skip ingest resources, add on search resources, add ingest resources, update search resources

Demo video, showing fine-grained provisioning happening on the search side. We provision an ingest pipeline & index, then on search, configure and provision a search pipeline separately, but keeping the existing ingest resources. Notice the search call made after running, and how data is still present. Lastly, shows that selecting different indices (if ingest is disabled) is persisted in the UI.

[screen-capture (21).webm](https://github.com/user-attachments/assets/28f72822-0090-4d0f-b116-9a368192dce1)

### Issues Resolved

Makes progress on #23 
Resolves #267

### Check List

### DIFF
--- a/public/pages/workflow_detail/workflow_inputs/workflow_inputs.tsx
+++ b/public/pages/workflow_detail/workflow_inputs/workflow_inputs.tsx
@@ -218,9 +218,9 @@ export function WorkflowInputs(props: WorkflowInputsProps) {
           // amount of time and may be long and expensive; we add this single
           // auto-fetching to cover the majority of provisioning updates which
           // are inexpensive and will finish within milliseconds.
-          new Promise((f) => setTimeout(f, 1000)).then(async () => {
+          setTimeout(async () => {
             dispatch(getWorkflow(updatedWorkflow.id as string));
-          });
+          }, 1000);
         })
         .catch((error: any) => {
           console.error('Error reprovisioning workflow: ', error);
@@ -254,9 +254,9 @@ export function WorkflowInputs(props: WorkflowInputsProps) {
                   // amount of time and may be long and expensive; we add this single
                   // auto-fetching to cover the majority of provisioning updates which
                   // are inexpensive and will finish within milliseconds.
-                  new Promise((f) => setTimeout(f, 1000)).then(async () => {
+                  setTimeout(async () => {
                     dispatch(getWorkflow(updatedWorkflow.id as string));
-                  });
+                  }, 1000);
                 })
                 .catch((error: any) => {
                   console.error('Error provisioning updated workflow: ', error);
@@ -289,13 +289,11 @@ export function WorkflowInputs(props: WorkflowInputsProps) {
     submitForm();
     await validateForm()
       .then(async (validationResults: { ingest?: {}; search?: {} }) => {
-        let relevantValidationResults = {} as { ingest?: {}; search?: {} };
-        if (includeIngest && validationResults.ingest !== undefined) {
-          relevantValidationResults.ingest = validationResults.ingest;
-        }
-        if (includeSearch && validationResults.search !== undefined) {
-          relevantValidationResults.search = validationResults.search;
-        }
+        const { ingest, search } = validationResults;
+        const relevantValidationResults = {
+          ...(includeIngest && ingest !== undefined ? { ingest } : {}),
+          ...(includeSearch && search !== undefined ? { search } : {}),
+        };
         if (Object.keys(relevantValidationResults).length > 0) {
           // TODO: may want to persist more fine-grained form validation (ingest vs. search)
           // For example, running an ingest should be possible, even with some
--- a/public/pages/workflow_detail/workflow_inputs/workflow_inputs.tsx
+++ b/public/pages/workflow_detail/workflow_inputs/workflow_inputs.tsx
@@ -5,7 +5,7 @@
 
 import React, { useCallback, useEffect, useState } from 'react';
 import { useSelector } from 'react-redux';
-import { useFormikContext } from 'formik';
+import { getIn, useFormikContext } from 'formik';
 import { debounce, isEmpty } from 'lodash';
 import {
   EuiButton,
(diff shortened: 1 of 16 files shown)


public/pages/workflow_detail/workflow_inputs/workflow_inputs.tsx
```diff
@@ -188,73 +192,111 @@ export function WorkflowInputs(props: WorkflowInputsProps) {
...
+                  console.error('Error provisioning updated workflow: ', error);
+                });
+            })
+            .catch((error: any) => {
+              console.error('Error updating workflow: ', error);
+            });
+        })
+        .catch((error: any) => {
+          console.error('Error deprovisioning workflow: ', error);
+        });
```
owaiskazi19, line 271: Can be combined under one catch block

```ts
catch (error: any) {
    console.error('Error during workflow update:', error);
    // Handle specific errors based on their source
    if (error.message.includes('deprovisionWorkflow')) {
      console.error('Error deprovisioning workflow:', error);
    } else if (error.message.includes('updateWorkflow')) {
      console.error('Error updating workflow:', error);
    } else if (error.message.includes('provisionWorkflow')) {
      console.error('Error provisioning updated workflow:', error);
    }
  }
```

public/pages/workflow_detail/workflow_inputs/workflow_inputs.tsx
```diff
@@ -188,73 +192,111 @@ export function WorkflowInputs(props: WorkflowInputsProps) {
...
+                  await sleep(1000);
+                  success = true;
+                  // Kicking off an async task to re-fetch the workflow details
+                  // after some amount of time. Provisioning will finish in an indeterminate
+                  // amount of time and may be long and expensive; we add this single
+                  // auto-fetching to cover the majority of provisioning updates which
+                  // are inexpensive and will finish within milliseconds.
+                  new Promise((f) => setTimeout(f, 1000)).then(async () => {
+                    dispatch(getWorkflow(updatedWorkflow.id as string));
+                  });
```
owaiskazi19, line 259:
```suggestion
                  setTimeout(async () => {
                     dispatch(getWorkflow(updatedWorkflow.id as string));
              }, 1000);
```

public/pages/workflow_detail/workflow_inputs/workflow_inputs.tsx
```diff
@@ -188,73 +192,111 @@ export function WorkflowInputs(props: WorkflowInputsProps) {
...
-      .then(async (validationResults: {}) => {
-        if (Object.keys(validationResults).length > 0) {
+      .then(async (validationResults: { ingest?: {}; search?: {} }) => {
+        let relevantValidationResults = {} as { ingest?: {}; search?: {} };
+        if (includeIngest && validationResults.ingest !== undefined) {
+          relevantValidationResults.ingest = validationResults.ingest;
+        }
+        if (includeSearch && validationResults.search !== undefined) {
+          relevantValidationResults.search = validationResults.search;
+        }
```
owaiskazi19, line 298: How about the below for L293-298

```ts
const { ingest, search } = validationResults;
relevantValidationResults = {
  ...(includeIngest && ingest !== undefined ? { ingest } : {}),
  ...(includeSearch && search !== undefined ? { search } : {}),
};
```

owaiskazi19 (commented):
LG overall with few styling comments

public/pages/workflow_detail/workflow_inputs/workflow_inputs.tsx
ohltyler: I'd actually prefer to leave as-is; less change for issues if the message schema or API names change etc. Also, we may have further logic in those `catch` blocks later on that is easier to handle within their standalone blocks.

public/pages/workflow_detail/workflow_inputs/workflow_inputs.tsx
ohltyler: nice. Will update manually in both places.

public/pages/workflow_detail/workflow_inputs/workflow_inputs.tsx
ohltyler: Lovely! Will update

owaiskazi19 (approved):

Issue #272, opened 2024-08-07
Make form state more granular


ohltyler:
Currently, we persist form state in redux, and update via `setDirty()`/`removeDirty()`. This is to enable/disable certain buttons in the stepped form flow (e.g., not going to search flow until ingest resources are provisioned). We persist this separately than the `Formik` form state, since we may do implicit validation/saving/submitting which would clean the form, but not necessarily submit and actually create resources.

Currently, the dirty state is across both search and ingest. Improve in such a way we have separate form state for each, giving us easier control over buttons being enabled/disabled, displaying more callouts, etc.

PR #273, opened 2024-08-07
Improve form usability & autosave

ohltyler:
### Description

Improves plugin usability in a few different ways:
- lower the autosave debounce amount from 10s -> 1s. Reason being, this minimizes the chances of updating a field, and it getting overwritten with some previous state. Long term, we may need to update how we handle autosave, as this could throttle or get expensive for making such often update workflow API calls. Worst case, we can fall back to manually saving.
- prevents unnecessary get workflow call after autosaving. the consequence is the "last updated" field is then not updated. Add TODO to handle at a later point.
- improved buttons to be disabled less often / more flexibility when navigating between forms
- gracefully handles running "run query" multiple times. The reprovision API will throw an error if there are no underlying template changes. We hide that error, and continue with executing the query.
- updates the presets - removes outdated ones, adds in a blank/custom one.
- adds TODOs for improving form state. Tracking issue w/ more details: #272 

### Issues Resolved
Makes progress on #23 

### Check List


public/pages/workflow_detail/workflow_inputs/workflow_inputs.tsx
```diff
@@ -676,6 +675,7 @@ export function WorkflowInputs(props: WorkflowInputsProps) {
                           onClick={() => {
                             validateAndRunIngestion();
                           }}
+                          // TODO: only enable if ingest is dirty
```
owaiskazi19, line 678: What do we mean here by "dirty"?

ohltyler: form is dirty / unsaved changes

owaiskazi19: Can we have a better name here? `Unsaved` or something

ohltyler: It's a common term for unsaved form data, and is used throughout `Formik` form library we use to maintain form and state. Prefer to leave as-is for consistency.

owaiskazi19: I see. Sounds good

owaiskazi19 (approved):

PR #293, opened 2024-08-16
Header Redesign

saimedhi:
### Description

- Updated the header to align with the new home page UI design.
- Users can access this new header by setting the useNewHomePage button to True.
- The old header remains the default option and is still available.
- Relocated the datasource picker mounting code from the workflow_detail.tsx file to the header.tsx file.
- Integrated the new Application Header for the workflow_detail page, with all header items now mounted onto the TopNavMenu.
- Added icons for "Export" and "Return to Projects" in the workflow_detail page.
- Fixed the breadcrumbs to be consistent with the new home page UI.

### Check List

  - [x] All tests pass
- [ x] Commits are signed per the DCO using `--signoff`


saimedhi:
[screen-capture (8).webm](https://github.com/user-attachments/assets/d5b552b8-17b5-43b4-a7fc-8af9c9dc25e5)

saimedhi:
@ohltyler, @jackiehanyang Please take a look.

public/pages/workflow_detail/components/header.tsx
```diff
@@ -52,14 +73,130 @@ export function WorkflowDetailHeader(props: WorkflowDetailHeaderProps) {
...
   }, [props.workflow]);
 
-  return (
+  const { TopNavMenu, HeaderControl } = getNavigationUI();
+  const { setAppRightControls } = getApplication();
+  const {
+    chrome: { setHeaderVariant },
+  } = getCore();
+  const uiSettings = getUISettings();
+  const showActionsInHeader = uiSettings.get('home:useNewHomePage');
```
ohltyler, line 82: nit: can we add this path as a constant in `common/constants`?

public/pages/workflow_detail/components/header.tsx
```diff
@@ -52,14 +73,130 @@ export function WorkflowDetailHeader(props: WorkflowDetailHeaderProps) {
...
+
```
ohltyler, line 83: super nit: can you add one-line comments on added `useEffect` hooks to describe what they're doing? especially since this is all pulled logic from core OSD, helps to have some context added if it changes later on.

ohltyler: also a little confused on what is happening in here, as it seems to run `setHeaderVariant` twice potentially.

public/pages/workflow_detail/components/header.tsx
```diff
@@ -52,14 +73,130 @@ export function WorkflowDetailHeader(props: WorkflowDetailHeaderProps) {
...
+      iconType: 'exit',
+      tooltip: 'Return to projects',
+      ariaLabel: 'Exit',
+      run: onExitButtonClick,
+      controlType: 'icon',
+    } as TopNavMenuIconData,
+  ];
+
+  const dataSourceEnabled = getDataSourceEnabled().enabled;
+  const dataSourceId = getDataSourceId();
```
ohltyler, line 124: For consistency, I would suggest moving these types of vars at the top of the fn, like around line 59. Same with the variables on lines 76-82. Then we can have all of those state vars etc in one place. Additionally, we want them defined before custom `useEffect` hooks typically, since the execution order matters. If these are defined/updated later, could cause weird state bugs.

public/pages/workflow_detail/components/header.tsx
```diff
@@ -52,14 +73,130 @@ export function WorkflowDetailHeader(props: WorkflowDetailHeaderProps) {
...
+  }
+
+  return showActionsInHeader ? (
+    <>
+      {isExportModalOpen && (
+        <ExportModal
+          workflow={props.workflow}
+          setIsExportModalOpen={setIsExportModalOpen}
+        />
+      )}
```
ohltyler, line 152: duplicate of lines 193-198. Since this state isn't dependent on `showActionsInHeader` being true or false, let's defined this outside of that first. Something like:
```
return ( 
<>
{ isExportModalOpen && (...) }
{ showActionsInHeader ? (<top nav menu stuff>) : (<existing header stuff>) }
</>
```

public/pages/workflow_detail/components/header.tsx
```diff
@@ -52,14 +73,130 @@ export function WorkflowDetailHeader(props: WorkflowDetailHeaderProps) {
...
+      <TopNavMenu
+        appName={PLUGIN_ID}
+        config={topNavConfig}
+        screenTitle={workflowName}
+        showDataSourceMenu={dataSourceEnabled ? true : false}
```
ohltyler, line 157: can be simplified to 
```
showDataSourceMenu={dataSourceEnabled}
```

public/pages/workflow_detail/workflow_detail.tsx
```diff
@@ -64,21 +58,28 @@ export function WorkflowDetail(props: WorkflowDetailProps) {
   const workflow = workflows[workflowId];
   const workflowName = workflow ? workflow.name : DEFAULT_NEW_WORKFLOW_NAME;
 
+  const {
+    chrome: { setBreadcrumbs },
+  } = getCore();
+  const uiSettings = getUISettings();
+  const showActionsInHeader = uiSettings.get('home:useNewHomePage');
```
ohltyler, line 65: nit: prefer to have as a constant

public/pages/workflow_detail/workflow_detail.tsx
```diff
@@ -64,21 +58,28 @@ export function WorkflowDetail(props: WorkflowDetailProps) {
...
       ]);
     } else {
-      getCore().chrome.setBreadcrumbs([
+      setBreadcrumbs([
         BREADCRUMBS.FLOW_FRAMEWORK,
-        BREADCRUMBS.WORKFLOWS(),
+        BREADCRUMBS.WORKFLOWS(dataSourceEnabled ? dataSourceId : undefined),
         { text: workflowName },
       ]);
     }
```
ohltyler, line 81: even better, I think this could be combined into just a single `setBreadcrumbs` call:
```
      setBreadcrumbs( showActionsInHeader ? [
        BREADCRUMBS.TITLE_WITH_REF(
          dataSourceEnabled ? dataSourceId : undefined
        ),
        BREADCRUMBS.WORKFLOW_NAME(workflowName),
        { text: '' },
      ] : [
        BREADCRUMBS.FLOW_FRAMEWORK,
        BREADCRUMBS.WORKFLOWS(dataSourceEnabled ? dataSourceId : undefined),
        { text: workflowName },
      ]);
```

public/pages/workflows/workflows.tsx
```diff
@@ -117,15 +132,12 @@ export function Workflows(props: WorkflowsProps) {
...
-      ]);
+    if (showActionsInHeader) {
+      setBreadcrumbs([BREADCRUMBS.TITLE]);
     } else {
-      getCore().chrome.setBreadcrumbs([
+      setBreadcrumbs([
         BREADCRUMBS.FLOW_FRAMEWORK,
-        BREADCRUMBS.WORKFLOWS(),
+        BREADCRUMBS.WORKFLOWS(dataSourceEnabled ? dataSourceId : undefined),
       ]);
```
ohltyler, line 141: similar as above, we can combine even more:
```
setBreadcrumbs( showActionsInHeader ? [BREADCRUMBS.TITLE] : [
        BREADCRUMBS.FLOW_FRAMEWORK,
        BREADCRUMBS.WORKFLOWS(dataSourceEnabled ? dataSourceId : undefined),
      ]);
```

public/store/reducers/models_reducer.ts
```diff
@@ -20,7 +20,7 @@ const SEARCH_MODELS_ACTION = `${MODELS_ACTION_PREFIX}/searchModels`;
 export const searchModels = createAsyncThunk(
   SEARCH_MODELS_ACTION,
   async (
-    { apiBody, dataSourceId }: { apiBody: {}, dataSourceId: string },
+    { apiBody, dataSourceId }: { apiBody: {}; dataSourceId?: string },
```
ohltyler, line 23: does this not cause errors?

public/utils/constants.ts
```diff
@@ -22,4 +22,12 @@ export const BREADCRUMBS = Object.freeze({
     text: 'Workflows',
     href: constructHrefWithDataSourceId(APP_PATH.WORKFLOWS, dataSourceId),
   }),
+  TITLE: { text: 'Search Studio' },
```
ohltyler, line 25: nit: prefer to add `Search Studio` as a constant.

public/pages/workflows/new_workflow/use_case.tsx
```diff
@@ -90,7 +90,11 @@ export function UseCase(props: UseCaseProps) {
                   .then((result) => {
                     const { workflow } = result;
                     history.replace(
-                      `${APP_PATH.WORKFLOWS}/${workflow.id}?dataSourceId=${dataSourceId}`
+                      `${APP_PATH.WORKFLOWS}/${workflow.id}${
+                        dataSourceId !== undefined
+                          ? `?dataSourceId=${dataSourceId}`
+                          : ''
+                      }`
```
ohltyler, line 97: can we utilize `constructHrefWithDataSourceId` instead? I see this repeated. Seems very similar. And if not, maybe we can make the helper fn more flexible to handle it, or a separate standalone fn.

ohltyler (commented):
This is awesome to see! thanks for the video as well. Just some mostly minor comments.

public/store/reducers/models_reducer.ts
saimedhi: I don't think this will cause any issue.

ohltyler: did it auto-correct to this? iirc there is compilation issues between using `,` and `;`. but if you're not facing issues, sounds fine

saimedhi:
**Overall testing**

[screen-capture (10).webm](https://github.com/user-attachments/assets/1ba2dce2-330d-40e8-adc6-a582e7b9aa62)

**Recorded after Fixing Close button** 

[screen-capture (12).webm](https://github.com/user-attachments/assets/01483339-f4b3-46e8-9664-28a7d94de9f5)

public/store/reducers/models_reducer.ts
saimedhi: - Yes, the formatter autocorrected. I didn't see any issues with using `;`
- I will remove this in the next PR, in case it needs to be rolled back.

public/pages/workflow_detail/components/header.tsx
```diff
@@ -79,33 +92,23 @@ export function WorkflowDetailHeader(props: WorkflowDetailHeaderProps) {
...
-    if (showActionsInHeader) {
+    if (SHOW_ACTIONS_IN_HEADER) {
       setHeaderVariant?.(HeaderVariant.APPLICATION);
     }
-
     return () => {
       setHeaderVariant?.();
     };
-  }, [setHeaderVariant, showActionsInHeader]);
+  }, [setHeaderVariant, SHOW_ACTIONS_IN_HEADER]);
```
ohltyler, line 103: Still bit confused on why it is returning some function on lines 100-102. Shouldn't this just be setting the header variant based on `SHOW_ACTIONS_IN_HEADER`? How about:
```
    if (SHOW_ACTIONS_IN_HEADER) {
      setHeaderVariant?.(HeaderVariant.APPLICATION);
    } else {
      setHeaderVariant?.()
    }
```

Or, do we even need to set it at all in the `else` case?

ohltyler: discussed offline, existing logic is fine, as the return fn is just the cleanup fn for when the component unmounts.

ohltyler (approved):

Issue #95, opened 2024-03-01
[META] Unit test coverage


ohltyler:
Tracking issue for adding unit tests for all pages and UI components
- [x] #103 
- [x] #105 
- [x] #96 
- [x] #97 
- [x] #98 
- [x] #99 
- [x] #100

PR #326, opened 2024-08-29
Added unit tests for WorkflowDetail component

saimedhi:
### Description

Added unit tests for WorkflowDetail component

### Issues Resolved

Related to #95 

### Check List

  - [x] All tests pass

### DIFF
--- a/public/pages/workflow_detail/workflow_detail.test.tsx
+++ b/public/pages/workflow_detail/workflow_detail.test.tsx
@@ -1,0 +1,50 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React from 'react';
+import { render } from '@testing-library/react';
+import { Provider } from 'react-redux';
+
+import {
+  BrowserRouter as Router,
+  RouteComponentProps,
+  Route,
+  Switch,
+} from 'react-router-dom';
+import { store } from '../../store';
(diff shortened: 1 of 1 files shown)


public/pages/workflow_detail/workflow_detail.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+        </Switch>
+      </Router>
+    </Provider>
+  );
+
+describe('WorkflowDetail', () => {
+  test('renders the page', () => {
+    const { container, baseElement } = renderWithRouter();
+    expect(baseElement.tagName).toBe('BODY');
+    expect(container.tagName).toBe('DIV');
```
ohltyler, line 48: This isn't a useful thing to test. We should test that certain wording exists, or other descriptors of components that we would be expecting when this renders. I would recommend looking at the AD OSD plugin unit tests for more examples. You can also refer to the [testing-library documentation](https://testing-library.com/docs/queries/about/) for more info and examples.

At a high-level, we use `jest` as the test framework, and `@testing-library/react` as a helper library for rendering and pulling out UI elements. A basic flow for testing UI components/pages is:
1. Render the component using `render()` (optionally with different props for different states)
2. Run `testing-library` queries (linked above) against that returned component to ensure certain elements exist etc.
3. Assert that those queries return what is expected using `jest`, such as `expect()`

PR #345, opened 2024-09-04
workflow_detail unit tests

saimedhi:
### Description

- Added workflow_detail unit tests. 
- Tests added for the following workflow types : "Custom", "Semantic search", "Hybrid search"

### Issues Resolved

Related to https://github.com/opensearch-project/dashboards-flow-framework/issues/95

### Check List

  - [x] All tests pass

### DIFF
--- a/public/pages/workflow_detail/workflow_detail.test.tsx
+++ b/public/pages/workflow_detail/workflow_detail.test.tsx
@@ -28,7 +28,13 @@ jest.mock('../../services', () => {
   };
 });
 
-const history = createMemoryHistory();
+const workflowId = '12345';
+const workflowName = 'test_workflow';
+
+const history = createMemoryHistory({
+  initialEntries: [`/workflow/${workflowId}`],
+});
+
 window.ResizeObserver = resizeObserverMock;
 
 const renderWithRouter = (
@@ -46,15 +52,11 @@ const renderWithRouter = (
               return <WorkflowDetail setActionMenu={jest.fn()} {...props} />;
             }}
           />
-          <Redirect from="/" to={`/workflow/${workflowId}`} />
         </Switch>
       </Router>
     </Provider>
   ),
 });
-
-const workflowId = '12345';
-const workflowName = 'test_workflow';
 
 describe('WorkflowDetail', () => {
   Object.values(WORKFLOW_TYPE).forEach((type) => {
--- a/public/pages/workflow_detail/workflow_detail.test.tsx
+++ b/public/pages/workflow_detail/workflow_detail.test.tsx
@@ -57,13 +57,7 @@ const workflowId = '12345';
 const workflowName = 'test_workflow';
 
 describe('WorkflowDetail', () => {
-  const workflowTypes = [
-    WORKFLOW_TYPE.CUSTOM,
-    WORKFLOW_TYPE.SEMANTIC_SEARCH,
-    WORKFLOW_TYPE.HYBRID_SEARCH,
-  ];
-
-  workflowTypes.forEach((type) => {
+  Object.values(WORKFLOW_TYPE).forEach((type) => {
     test(`renders the page with ${type} type`, () => {
       const { getAllByText, getByText, getByRole } = renderWithRouter(
         workflowId,
--- a/public/pages/workflow_detail/workflow_detail.test.tsx
+++ b/public/pages/workflow_detail/workflow_detail.test.tsx
@@ -1,0 +1,71 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React from 'react';
+import { render } from '@testing-library/react';
+import { Provider } from 'react-redux';
+import {
+  RouteComponentProps,
+  Route,
+  Switch,
+  MemoryRouter,
+} from 'react-router-dom';
+import { WorkflowDetail } from './workflow_detail';
+import { WorkflowDetailRouterProps } from '../../pages';
+import '@testing-library/jest-dom';
+import { mockStore } from '../../../test/utils';
+
+jest.mock('../../services', () => {
+  const { mockCoreServices } = require('../../../test');
+  return {
+    ...jest.requireActual('../../services'),
+    ...mockCoreServices,
+  };
+});
+
+const renderWithRouter = (initialEntries: string[]) =>
+  render(
+    <Provider store={mockStore}>
+      <MemoryRouter initialEntries={initialEntries}>
+        <Switch>
+          <Route
+            path="/workflow/:workflowId"
+            render={(props: RouteComponentProps<WorkflowDetailRouterProps>) => (
+              <WorkflowDetail setActionMenu={jest.fn()} {...props} />
+            )}
+          />
+        </Switch>
+      </MemoryRouter>
+    </Provider>
+  );
+
+describe('WorkflowDetail', () => {
+  test('renders the page with workflowId parameter', () => {
+    const workflowId = '12345';
+    const { getAllByText, getByText, getByRole } = renderWithRouter([
+      `/workflow/${workflowId}`,
+    ]);
+
+    expect(getAllByText('test_workflow').length).toBeGreaterThan(0);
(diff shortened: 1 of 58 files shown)


minalsha (approved):

public/pages/workflow_detail/workflow_detail.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+  return {
+    ...jest.requireActual('../../services'),
+    ...mockCoreServices,
+  };
+});
+
+const renderWithRouter = (initialEntries: string[]) =>
+  render(
+    <Provider store={mockStore}>
+      <MemoryRouter initialEntries={initialEntries}>
```
ohltyler, line 31: How does `MemoryRouter` work? What about doing something similar to AD, where we can pass in some limited info (resource ID for example), and construct the route inline? See https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/pages/DetectorDetail/containers/__tests__/DetectorDetail.test.tsx#L33:L39 as an example.

public/pages/workflow_detail/workflow_detail.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+  );
+
+describe('WorkflowDetail', () => {
+  test('renders the page with workflowId parameter', () => {
+    const workflowId = '12345';
+    const { getAllByText, getByText, getByRole } = renderWithRouter([
+      `/workflow/${workflowId}`,
+    ]);
+
+    expect(getAllByText('test_workflow').length).toBeGreaterThan(0);
```
ohltyler, line 51: Let's move things like IDs / names into constants instead of hardcoding. That makes it easier to reuse in other tests.

test/utils.ts
```diff
@@ -0,0 +1,80 @@
...
+export const mockStore = {
+  getState: () => ({
+    opensearch: {
+      errorMessage: '',
+    },
+    workflows: {
+      loading: false,
+      errorMessage: '',
+      workflows: {
+        '12345': {
```
ohltyler, line 15: Suggest to move this entire workflow JSON into a constant. Or better yet, a util fn to generate a dummy workflow based on limited input. For example,
```
function generateWorkflow(workflowId: string, workflowName: string): Workflow
```

ohltyler: Eventually these types of functions can get more complex, take in more parameters, etc. For example, workflows with multiple processors, ingest disabled, different index settings, etc.

public/pages/workflow_detail/workflow_detail.test.tsx
saimedhi: - Using MemoryRouter is better for tests because it lets you simulate routing in memory without needing a real URL, making tests easier and more controlled.
- https://v5.reactrouter.com/web/api/MemoryRouter

saimedhi: - Alternate option is to  use`Router instead of MemoryRouter with  history = createMemoryHistory(); ` Will try this option and raise a PR. Thank you.

ohltyler: Is there a use case we want/need to persist history? I'm only proposing to keep it simple how we have other tests and just use `Router`. We still are passing the url path on line 34 regardless.

test/utils.ts
```diff
@@ -0,0 +1,80 @@
...
+  subscribe: jest.fn(),
+  replaceReducer: jest.fn(),
+  [Symbol.observable]: jest.fn(),
+};
+
+global.ResizeObserver = class {
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+};
```
ohltyler, line 80: just to double check, since I'm not super familiar - does the `global.*` create any issues when running code in public/ or server/? Want to make sure this doesn't interfere / override any functionality when the plugin is running outside of the test env.

saimedhi: Replacing it with below mock. 
```
const resizeObserverMock = jest.fn().mockImplementation(() => ({
  observe: jest.fn(),
  unobserve: jest.fn(),
  disconnect: jest.fn(),
}));

window.ResizeObserver = resizeObserverMock;

```

public/pages/workflow_detail/workflow_detail.test.tsx
saimedhi: - @ohltyler,  We need history while rendering header component . https://github.com/saimedhi/dashboards-flow-framework/blob/main/public/pages/workflow_detail/components/header.tsx#L58. 

- In such cases, AD is using `createMemoryHistory();` in test cases. Let me know if you feel something is incorrect. 

- If history is not given getting below errors. 
```

The prop `history` is marked as required in `Router`, but its value is `undefined`.

Uncaught [TypeError: Cannot read properties of undefined (reading 'location')] 
          
          
```

ohltyler: got it! That makes sense

test/utils.ts
```diff
@@ -0,0 +1,274 @@
...
+function generateWorkflow(
+  workflowId: string,
+  workflowName: string,
+  workflowType: WORKFLOW_TYPE
+): Workflow {
+  return {
+    id: workflowId,
+    name: workflowName,
+    version: { template: '1.0.0', compatibility: ['2.17.0', '3.0.0'] },
+    ui_metadata: {
```
ohltyler, line 52: The schema & default values inside of `ui_metadata` are still changing, and will definitely change more in the future. For consistency and maintainability, how about we leverage the helper fns here? https://github.com/opensearch-project/dashboards-flow-framework/blob/main/public/pages/workflows/new_workflow/utils.ts

This is the source of truth for the default configs. If we reuse it in tests, we can guarantee the tests and UI would be consuming the same values.

Still worth having the params like id/name/type here. Maybe can use the existing fns for generating all of the metadata, and then overwrite a few things like id/name based on the params.

public/pages/workflow_detail/workflow_detail.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+describe('WorkflowDetail', () => {
+  const workflowTypes = [
+    WORKFLOW_TYPE.CUSTOM,
+    WORKFLOW_TYPE.SEMANTIC_SEARCH,
+    WORKFLOW_TYPE.HYBRID_SEARCH,
+  ];
+
+  workflowTypes.forEach((type) => {
+    test(`renders the page with ${type} type`, () => {
```
ohltyler, line 67: Rather than re-defining the list of types here, you can just use the enum directly.
something like:
```
Object.values(WORKFLOW_TYPE).forEach((type) => ...)
```
I like the idea of automatically generating the test cases for each workflow type, nice!

test/utils.ts
ohltyler: It's my bad for not showing you that helper fn before. Hopefully that makes more sense and gives you some more understanding on how we generate the configs in general :)

public/pages/workflow_detail/workflow_detail.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+    <Provider store={mockStore(workflowId, workflowName, workflowType)}>
+      <Router history={history}>
+        <Switch>
+          <Route
+            path="/workflow/:workflowId"
+            render={(props: RouteComponentProps<WorkflowDetailRouterProps>) => {
+              return <WorkflowDetail setActionMenu={jest.fn()} {...props} />;
+            }}
+          />
+          <Redirect from="/" to={`/workflow/${workflowId}`} />
```
ohltyler, line 49: wondering why we need this redirect? Can't we specify the path directly on line 44?

ohltyler (approved):
These can be covered in a followup PR

public/pages/workflow_detail/workflow_detail.test.tsx
saimedhi: - @ohltyler, Removed **Redirect** and instead adding **initialEntries** within the history. 
- And also below path should not be removed. Whenever URL matches the below pattern then workflowId will be dynamically obtained from URL. Similar to the actual use case of workflowDetail Component. 
`path="/workflow/:workflowId"`

ohltyler (approved):

PR #360, opened 2024-09-10
Improving and expanding functionality checks for the workflow_detail page

saimedhi:
### Description

Improving and expanding functionality checks for the workflow_detail page. Added tests for below functionalities. 
- The "Create an ingest pipeline" option is selected by default
- The "Skip ingestion pipeline" option is disabled by default
- The "Run ingestion" button is enabled by default
- The "Search pipeline >" button is disabled by default
- Clicking the "close" button goes back to the list page
- Clicking the "export" button opens the export component
- Collapsing the "Tools" panels hides those components
- The default components in the ReactFlow workspace are visible. Default- Visual. 

### Issues Resolved

Related to #95 

### Check List

  - [x] All tests pass


public/pages/workflow_detail/workflow_detail.test.tsx
```diff
@@ -82,6 +92,86 @@ describe('WorkflowDetail', () => {
...
+      expect(jsonButton).toHaveClass('euiFilterButton-hasActiveFilters');
+
+      // Testing collapsible Tools panel
+      // Get the initial state of the Tools panel
+      const toolsPanelInitially = container.querySelector('#tools_panel_id');
+      expect(toolsPanelInitially).toBeVisible();
+      // Find and click the button to adjust panel sizes
+      const adjustPanelSizeButton = container.querySelector(
+        'button[aria-label="Press up or down to adjust panels size"].euiResizableButton--vertical'
+      );
```
ohltyler, line 146: Overall this is awesome to see, thanks for all the additions! One suggestion for these types of query selection, is they can be more prone to break over time - for example, class names, internal text, underlying components, etc. may change due to changes in this codebase, or from the upstream UI libraries. One thing we can do is use the common pattern of applying a `data-test-subj` id to components we will test later on, that we have a static reference to a component that we have full control over.

For example, see the data-test-subj set [here](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/pages/DetectorsList/components/EmptyMessage/EmptyMessage.tsx#L42)
And referenced in the test [here](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/pages/DetectorsList/components/EmptyMessage/__tests__/EmptyMessage.test.tsx#L72)

I'd recommend seeing if some of the components pulled in here can be replaced with these types of references. Understandably, not all may be able to, as we have to check for class changes like `-isCollapsed` like you have on line 161.

ohltyler (approved):
LGTM!

PR #373, opened 2024-09-12
Standardize source data to input/output transform modals

ohltyler:
### Description
This PR standardizes the generated source input to consistently match exactly the way the ML processors will transform the inputs. For `full_response_path=false/true`, both are displayed exactly how the ML processor will parse it. The one caveat is, there is still work to be done on the search response side to handle many-to-one/one-to-one toggling better.

Also has other small fixes:
- handles when input or output is missing from the model interface (both are not required)
- handles empty search pipeline edge case when simulating source output
- triggers `touched` when toggling boolean field - before, changing these values was not triggering save to be available.
- removes the invalid check of bracket notation not available for the input/output map configurations
- adds loading logic when fetching to prevent overloading potentially expensive calls (simulation on ingest/search pipelines can be using ML credits & have large latency)
- fixes bug of showing multi-predict selector when input map is empty

Demo video, showing the updated input transforms and output transform source data for ingest/search req. Extra data, arrays, etc. have all been removed to consistently show the exact input JSON to the ML processor inputs, including when full_response_path is true/false

[screen-capture (19).webm](https://github.com/user-attachments/assets/6fd26626-3a8e-4b9d-9cf7-e369fcd63b07)

### Check List

### DIFF
--- a/common/interfaces.ts
+++ b/common/interfaces.ts
@@ -394,8 +394,8 @@ export type ModelInputFormField = ModelInput & {
 export type ModelOutputFormField = ModelInputFormField;
 
 export type ModelInterface = {
-  input: ModelInput;
-  output: ModelOutput;
+  input?: ModelInput;
+  output?: ModelOutput;
 };
 
 export type ConnectorParameters = {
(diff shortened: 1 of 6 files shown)


saimedhi (approved):
LGTM

PR #385, opened 2024-09-17
Support relevant form toggling in advanced transform modals

### LINKED ISSUE #367: Support many-to-one in output transform modals for ML search response processors
ML search response processors uniquely allow to process one document at a time, or all at once (see one_to_one parameter in the docs [here](https://opensearch.org/docs/latest/search-plugins/search-pipelines/ml-inference-search-response/#request-fields)). We should customize the fetched data, and how the transforms work, depending on if one_to_one is true/false. Currently, it is showing as one-to-one, as that is the only experience for the other 2 ingest and search request contexts.

ohltyler:
### Description

This PR adds specific, relevant forms to be exposed in the advanced transform modals, such that users can toggle and test directly within the modal, instead of navigating to/from the advanced settings fields. Specifically:
- exposes `one_to_one` toggling on INPUT transform for SEARCH RESPONSE only. This changes what the source input will be (single value vs. array of values)
- exposes `full_response_path` toggling on OUTPUT transforms for INGEST / SEARCH RESPONSE only. This changes what the source output will be (nested model response vs. full model response)

More details
- adds visual support for the default many-to-one behavior on input transform (when `one_to_one` is false). In this scenario, we collapse the values from all of the documents into a single list, to pass as a single parameter for one inference call via the ML processor
- updates `generateTransform` to support the above use case ^
- exposes `override` field for search resp processor
- limits input docs to the input transform, to be 10
- resets form inputs when form values are toggled

Demo video, showing `one_to_one` toggling on the input transform, and `full_response_path` toggling on the output transform. Note the model input selector in the output field changes from a selector when `false`, to a freeform text when `true`. This is because when it is `true`, the full model response path is passed, and users must write custom JSON path to parse out the relevant values they need.

[screen-capture (24).webm](https://github.com/user-attachments/assets/dac4547f-66d9-463e-96c5-34982b7728c8)

### Issues Resolved

Resolves #367 

### Check List

### DIFF
--- a/common/constants.ts
+++ b/common/constants.ts
@@ -155,6 +155,8 @@ export enum COMPONENT_CLASS {
  */
 export const ML_INFERENCE_DOCS_LINK =
   'https://opensearch.org/docs/latest/ingest-pipelines/processors/ml-inference/#configuration-parameters';
+export const ML_INFERENCE_RESPONSE_DOCS_LINK =
+  'https://opensearch.org/docs/latest/search-plugins/search-pipelines/ml-inference-search-response/#request-fields';
 export const ML_CHOOSE_MODEL_LINK =
   'https://opensearch.org/docs/latest/ml-commons-plugin/integrating-ml-models/#choosing-a-model';
 export const TEXT_CHUNKING_PROCESSOR_LINK =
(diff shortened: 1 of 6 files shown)


saimedhi (approved):
Code Changes Looks good to me. From the UI, one_to_one toggling on the input transform, and full_response_path toggling on the output transform seem to work rightly. Thank you, Tyler.

PR #388, opened 2024-09-18
Expanded functionality tests for New Workflow and Workflow List components.

saimedhi:
### Description

- Expanded functionality tests for the New Workflow and Workflow List components.
- In the Workflow List component, tested sorting, pagination, delete actions, and view resources.
- In the New Workflow page, tested templates and quick configuration.
- Will raise a follow-up PR to test the search bar in both of the above-mentioned components.

### Check List

  - [x] All tests pass

### DIFF
--- a/public/pages/workflows/new_workflow/new_workflow.test.tsx
+++ b/public/pages/workflows/new_workflow/new_workflow.test.tsx
@@ -12,7 +12,7 @@ import userEvent from '@testing-library/user-event';
 import configureStore from 'redux-mock-store';
 import * as ReactReduxHooks from '../../../store/store';
 import '@testing-library/jest-dom';
-import { loadPresetTemplates } from '../../../../test/utils';
+import { loadPresetWorkflowTemplates } from '../../../../test/utils';
 import { INITIAL_ML_STATE } from '../../../../public/store';
 
 jest.mock('../../../services', () => {
@@ -28,7 +28,7 @@ const initialState = {
   ml: INITIAL_ML_STATE,
   presets: {
     loading: false,
-    presetWorkflows: loadPresetTemplates(),
+    presetWorkflows: loadPresetWorkflowTemplates(),
   },
 };
 const store = mockStore(initialState);
@@ -59,6 +59,8 @@ describe('NewWorkflow', () => {
     expect(getAllByText('Hybrid Search')).toHaveLength(1);
     expect(getAllByText('Multimodal Search')).toHaveLength(1);
     expect(getAllByText('Semantic Search')).toHaveLength(1);
+    expect(getAllByText('Retrieval-augmented generation')).toHaveLength(1);
+    expect(getAllByText('Sentiment analysis')).toHaveLength(1);
   });
 
   test('renders the quick configure for preset workflow templates', async () => {
--- a/public/pages/workflows/new_workflow/new_workflow.test.tsx
+++ b/public/pages/workflows/new_workflow/new_workflow.test.tsx
@@ -59,8 +59,8 @@ describe('NewWorkflow', () => {
     expect(getAllByText('Hybrid Search')).toHaveLength(1);
     expect(getAllByText('Multimodal Search')).toHaveLength(1);
     expect(getAllByText('Semantic Search')).toHaveLength(1);
-    expect(getAllByText('Retrieval-augmented generation')).toHaveLength(1);
-    expect(getAllByText('Sentiment analysis')).toHaveLength(1);
+    expect(getAllByText('Retrieval-Augmented Generation')).toHaveLength(1);
+    expect(getAllByText('Sentiment Analysis')).toHaveLength(1);
   });
 
   test('renders the quick configure for preset workflow templates', async () => {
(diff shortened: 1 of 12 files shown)


public/pages/workflows/workflow_list/workflow_list.test.tsx
```diff
@@ -33,5 +64,105 @@ describe('WorkflowList', () => {
...
+    expect(sortButtons[1]).toBeInTheDocument();
+    userEvent.click(sortButtons[1]!);
+    await waitFor(() => {
+      expect(getAllByText('Custom').length).toBeGreaterThan(0);
+      expect(queryByText('Unknown')).toBeNull();
+    });
+    userEvent.click(sortButtons[1]!);
+    await waitFor(() => {
+      expect(queryByText('Unknown')).toBeNull();
+      expect(getAllByText('Custom').length).toBeGreaterThan(0);
```
saimedhi, line 107: @ohltyler, Added  **waitFor** for the assertion instead of userEvent.click after some research. 
- In the above case, It ensures that the component has re-rendered after the click event, and the assertion is made only when the expected text is visible.

- On the other hand, When added to **userEvent.click** , The waitFor is waiting for the click event to happen, but it doesn't wait for the component to re-render after the click event.

ohltyler: I see, so it's essentially re-running the assertion until some default timeout, or it passes. Which in our case, should pass, once the click has happened and everything has been re-rendered. makes sense, thanks!

public/pages/workflows/new_workflow/new_workflow.test.tsx
```diff
@@ -31,8 +47,49 @@ const renderWithRouter = () =>
   );
 
 describe('NewWorkflow', () => {
-  test('renders the search bar', () => {
-    const { getByPlaceholderText } = renderWithRouter();
+  beforeEach(() => {
+    jest.clearAllMocks();
+    jest.spyOn(ReactReduxHooks, 'useAppDispatch').mockReturnValue(mockDispatch);
+  });
```
ohltyler, line 53: nice. i imagine we'll have this as a common pattern for tests.

public/pages/workflows/new_workflow/new_workflow.test.tsx
```diff
@@ -31,8 +47,49 @@ const renderWithRouter = () =>
...
+
+  test('renders the preset workflow templates', () => {
+    const { getByPlaceholderText, getAllByText } = renderWithRouter();
     expect(getByPlaceholderText('Search')).toBeInTheDocument();
+    expect(getAllByText('Custom')).toHaveLength(1);
+    expect(getAllByText('Hybrid Search')).toHaveLength(1);
+    expect(getAllByText('Multimodal Search')).toHaveLength(1);
+    expect(getAllByText('Semantic Search')).toHaveLength(1);
+    expect(getAllByText('Retrieval-Augmented Generation')).toHaveLength(1);
+    expect(getAllByText('Sentiment Analysis')).toHaveLength(1);
```
ohltyler, line 63: for consistency, let's leverage the `WORKFLOW_TYPE` enum. In fact, we should be able to iterate through them in one loop (besides `WORKFLOW_TYPE.UNKNOWN`). Can do something like:

```
Object.values(WORKFLOW_TYPE).forEach((type) => {
  if (type !== WORKFLOW_TYPE.UNKNOWN) {
     expect(getAllByText(type)).toHaveLength(1)
  }
})
```

ohltyler: Then as we add more, they will be covered in this test case.

public/pages/workflows/workflow_list/workflow_list.test.tsx
```diff
@@ -18,6 +22,33 @@ jest.mock('../../../services', () => {
...
+}));
+const workflowSet: [
+  string,
+  string,
+  WORKFLOW_TYPE
+][] = workflows.map(({ id, name, type }): [string, string, WORKFLOW_TYPE] => [
+  id,
+  name,
+  type,
+]);
```
ohltyler, line 40: I don't think this is needed? `workflows` is already producing an array of objs containing `id, name, type`. Also, I'm seeing the pattern `id, name, type` repeated several times, usually a good sign to have a constant. How about having a constant somewhere that can be referenced instead? something like:
```
type TestWorkflowInput {
  id: string,
  name: string,
  type: WORKFLOW_TYPE
}
```

test/utils.ts
```diff
@@ -17,25 +17,30 @@ import {
...
-          [workflowId]: generateWorkflow(
-            workflowId,
-            workflowName,
-            workflowType
-          ),
-        },
+        workflows: workflowSets.reduce(
+          (
+            acc: Record<string, Workflow>,
+            [workflowId, workflowName, workflowType]
```
ohltyler, line 33: same as previous comment. Prefer to reference some type/interface here instead.

ohltyler (approved):
LGTM!

PR #389, opened 2024-09-18
Add dedicated prompt configuration for ML search resp processors

### LINKED ISSUE #380: Increase support for prompt template building and transformation
A common user journey for a RAG use case is constructing a dynamic prompt template that takes input on-the-fly to pass to an LLM model (e.g., "summarize these found documents: ..."). Currently, users have to manually build the prompt templates from scratch, set as a sub-value in the model config, and blindly run to see the results. We should add a few enhancements:

- [x] view the prompt template and injected values when constructing the input transforms #379 
- [ ] more presets and/or library of available prompt templates to select from, and a simpler way to configure

ohltyler:
### Description

Adds a dedicated section and modal for optionally configuring a prompt. Only exposing in the search response context, as that is the most likely place users will be setting it up. Can easily be enabled for other contexts if wanted later on. The dedicated modal has 3 main parts: 1/ dropdown with prompt template presets for users to select from, 2/ the user-editable text editor that displays the prompt, and 3/ the list of model inputs (and corresponding placeholder strings) that users can copy/paste, in order to insert these dynamic values into the template at runtime (this part is omitted if no input interface found). A common example of this, is passing returned documents as context in the prompt template for some LLM to answer a question with / summarize/ etc.

Implementation details:
- adds prompt presets in constants. just 2 very basic ones for now
- updates `MLProcessorInputs` to include a new section to configure the prompt and render the modal if clicked
- new `ConfigurePromptModal` component to handle the user interactions and configuration as described above. any updates to the prompt will be reflected in the form, by updating the `prompt` sub-field under the model config form field.
- minor: refactors the new modal and existing input/output modals into a standalone `modals` module
- minor: adds a `touched` check when selecting a new query, so the save button is enabled immediately when changed (if nothing else is dirty)
- minor: adds index config injection in the quick configure modal for RAG use case, so any configured text field, is set as a `text` field index mapping by default

Demo video, showing a new RAG use case with no prompt template. Shows configuration of the template using a preset, also including the dynamic `context` input to be injected into the template. The full example with injected inputs is available to view when configuring the input transform. When running, the results are as expected, containing a readable summary of the results found.

[screen-capture (1).webm](https://github.com/user-attachments/assets/ad547aed-2a11-4e0b-b137-8fe188f3b8c4)

### Issues Resolved
Resolves #380 

### Check List

### DIFF
--- a/public/pages/workflow_detail/workflow_inputs/processor_inputs/modals/configure_prompt_modal.tsx
+++ b/public/pages/workflow_detail/workflow_inputs/processor_inputs/modals/configure_prompt_modal.tsx
@@ -1,0 +1,184 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import React, { useState, useEffect } from 'react';
+import { useFormikContext, getIn } from 'formik';
+import { cloneDeep, isEmpty, set } from 'lodash';
+import {
+  EuiCodeEditor,
+  EuiFlexGroup,
+  EuiFlexItem,
+  EuiModal,
+  EuiModalBody,
+  EuiModalFooter,
+  EuiModalHeader,
+  EuiModalHeaderTitle,
+  EuiCompressedSelect,
+  EuiSelectOption,
+  EuiSmallButton,
+  EuiSpacer,
+  EuiText,
+  EuiPopover,
+  EuiSmallButtonEmpty,
+  EuiPopoverTitle,
+  EuiCodeBlock,
+  EuiCallOut,
+  EuiCode,
+} from '@elastic/eui';
+import {
+  IConfigField,
+  IProcessorConfig,
+  IngestPipelineConfig,
+  JSONPATH_ROOT_SELECTOR,
+  ML_INFERENCE_DOCS_LINK,
+  ML_INFERENCE_RESPONSE_DOCS_LINK,
+  MapArrayFormValue,
+  ModelInterface,
+  PROCESSOR_CONTEXT,
+  SearchHit,
+  SearchPipelineConfig,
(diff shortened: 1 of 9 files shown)


public/pages/workflow_detail/workflow_inputs/processor_inputs/modals/configure_prompt_modal.tsx
```diff
@@ -0,0 +1,296 @@
...
+      values,
+      `${props.baseConfigPath}.${props.config.id}.model_config`
+    ) as string;
+    try {
+      const prompt = JSON.parse(modelConfigString)?.prompt;
+      if (!isEmpty(prompt)) {
+        setPromptStr(prompt);
+      } else {
+        setPromptStr('');
+      }
```
saimedhi, line 87: @ohltyler, shall we change it to below code?
```
const prompt = JSON.parse(modelConfigString)?.prompt ?? '';
setPromptStr(prompt);
```

saimedhi (approved):
Thank you Tyler, Looks good to me.

public/pages/workflow_detail/workflow_inputs/processor_inputs/modals/configure_prompt_modal.tsx
ohltyler: [isEmpty](https://www.geeksforgeeks.org/lodash-_-isempty-method/) handles undefined, which is what is returned if no `prompt` field is found, so we don't need to add a fallback to convert to empty string.
```
Lodash _.isEmpty() checks if a value is considered empty. This includes objects with no own properties, arrays with no elements, strings of length zero, and other values like null, undefined, and false.
```

Issue #109, opened 2024-03-06
Re-enable lint workflow


ohltyler:
#108 disabled the lint workflow to have ongoing development still have fully passing PRs and tracking main/2.x branches. Before this is released and before the plugin becomes more stable, this should be re-enabled to enforce lint checking on PRs and merges to main/2.x.

ohltyler:
Setting 2.15 as a target goal to have this workflow re-enabled.

ohltyler:
Additionally, `pre-commit` configuration should be added back, to enable linting on every commit by default. It was removed due to lint violations during rapid development.

In `package.json`, add:
```
  "pre-commit": [
    "lint:es:precommit"
  ],
```

PR #401, opened 2024-09-25
Add Codecov for test coverage reporting

saimedhi:
### Description

- Add Codecov for test coverage reporting.

### Issues Resolved
Related to #109

### Check List

  - [x] All tests pass

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -102,10 +102,3 @@ jobs:
         run: |
           cd OpenSearch-Dashboards/plugins/dashboards-flow-framework
           yarn run test:jest --coverage
-      - name: Uploads coverage
-        uses: codecov/codecov-action@v3
-        with:
-          verbose: true 
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -45,6 +45,12 @@ jobs:
           su `id -un 1000` -c "source $NVM_DIR/nvm.sh && nvm use && node -v && yarn -v &&
                                cd ./plugins/dashboards-flow-framework &&
                                whoami && yarn osd bootstrap && yarn build && yarn run test:jest --coverage"
+      - name: Uploads coverage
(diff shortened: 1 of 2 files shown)


saimedhi:
Codecov token added is not getting detected by the workflows and throwing errors. Few other repos are experiencing similar issue. Raised bug report in codecov library [here](https://github.com/codecov/feedback/issues/525).

saimedhi:
- @ohltyler, please take a look at this Pr and merge it if possible. 
- From the conversation here, noticed that codecov workflow maynot succeed when Pull Request is raised but codecov update checks will succeed when  PR is merged.

ohltyler:
Few questions:
1. What is the reasoning for the extra params in the workflow step? In AD plugin, we just use the `codecov-action` ([ref](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/.github/workflows/build-and-test-workflow.yml#L48))
2. Suggest we just add this once for one env (linux), instead of adding the step multiple times to prevent duplicate uploads. Or, even better, add as a standalone github workflow file that just handles the coverage so it is decoupled from this 'build and test' workflow. WDYT?

saimedhi:
> Few questions:
> 
> 1. What is the reasoning for the extra params in the workflow step? In AD plugin, we just use the `codecov-action` ([ref](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/.github/workflows/build-and-test-workflow.yml#L48))
> 2. Suggest we just add this once for one env (linux), instead of adding the step multiple times to prevent duplicate uploads. Or, even better, add as a standalone github workflow file that just handles the coverage so it is decoupled from this 'build and test' workflow. WDYT?

1. Added codecov token, which is optional but good to have. 
2. I agree that we should have codecov uploading step running only once. As in the [codecov](https://app.codecov.io/gh/opensearch-project/dashboards-flow-framework), I didn't see any platform specific codecov metrics. Made the change.

ohltyler (approved):
LGTM thanks for adding!

saimedhi:
- Codecov successfully updated for main, 2.x branches 
- Current code coverage is **48.44%** as seen in below link. 

https://app.codecov.io/gh/opensearch-project/dashboards-flow-framework/tree/main

Issue #393, opened 2024-09-19
Make `prompt` parameter field more flexible


ohltyler:
Currently, the default prompt configuration in the model config is called `prompt`. But, depending on the connector details and/or model interfaces, the field could be called something other than `prompt`, or in some different format than just `string`. To support these other use cases easier, we should make the hardcoded logic more generic, but still having sufficient presets available.

ohltyler:
Marking as complete for now, abstracted into a constant via #403. Remaining flexibility work will all be dependent on the preset connectors and having a standard pattern, which we do not have for now. When that is standardized, we can revisit if this needs to be updated or more flexible to handle different ways prompts/inputs to LLMs are handled.

PR #403, opened 2024-10-01
Improve prompt config UX; support JSONPath bug fix

ohltyler:
### Description

Misc improvements PR:
- improves the logic on when to expose the "Configure prompt" section in the ML processor configs. Rather than exposing by default for ML search response processors, instead only expose if there is a present 'prompt` parameters via the connector blueprint, regardless of context (ingest,search req, search resp). In general, this will require some consistency on the expected blueprints to contain such field for such use cases - ongoing discussions with ML team to sort this part out.
- supports / is consistent with the bug fix in ML ingest processors, (see [here](https://github.com/opensearch-project/ml-commons/pull/2985))such that we do not always default to an array obj. Updates the underlying `jsonpath` call from `query` -> `value`, where `value` returns the first hit in the target obj, instead of wrapping all hits in an arr.
- Minor wording improvements of the 'Tools' tabs
- changing hardcoded `prompt` field to a constant. we may have more, and/or have a regex for determining when a connector has a configurable "prompt"/"prompt template" etc. (tracked via #393)

### Check List

### DIFF
--- a/common/constants.ts
+++ b/common/constants.ts
@@ -480,3 +480,4 @@ export const EMPTY_MAP_ENTRY = { key: '', value: '' } as MapEntry;
 export const MODEL_OUTPUT_SCHEMA_NESTED_PATH =
   'output.properties.inference_results.items.properties.output.items.properties.dataAsMap.properties';
 export const MODEL_OUTPUT_SCHEMA_FULL_PATH = 'output.properties';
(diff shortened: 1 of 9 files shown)


saimedhi (approved):

saimedhi:
Hello @ohltyler, would it be possible to provide a screen recording showcasing any changes from the user's perspective?

PR #407, opened 2024-10-02
Add dedicated UX for query template building (ML search req processors)

### LINKED ISSUE #404: Improve query template building experience
Currently, the only way to override a query in search request processors is via `query_template` field nested under advanced settings. There is no further help, such as presets, the placeholders and what they mean, etc.

Screenshot 2024-10-01 at 3 34 34 PM

Ideally, we have a more tailored experience, similar to prompt building, which can have presets, present the available parameters that could be injected, and provide more helper text in general regarding why it exists:

Screenshot 2024-10-01 at 3 37 07 PM

ohltyler:
### Description

This PR improves the UX for query configuration/overriding the query via a `query_template` field exposed in the ML search request processors. This is a common pattern for vector search use cases, to generate a valid knn query, injecting the returned vector embedding directly into the new query. Before, this was nested under 'Advanced settings', under 'Query template'. Now, we add a dedicated modal similar to prompt-building for this particular scenario.

More details
- adds a 'Override query' option for ML search req processors only
- supports preset selection, the same as the base query configuration (one difference, is once a preset is selected, we sanitize to be in a proper templated format, since the value here is a _template_ rather than a _string_)
- similar to the prompt modal, we expose some model configuration. In this case, the model outputs. Note we dynamically generate the list based on the presence of a configured output map or not. If there is an output map, then pull the keys (the transformed/new outputs) as available options to inject in the template. If there is no output map, then pull the model outputs from the interface directly, as this implies no further transformation is needed
- minor: adds some helper text in the configure prompt modal
- bug fix of back button disabled when skipping ingest altogether

Demo video, showing the new modal, and the various options shown under "Model outputs" based on any output map configuration.

[screen-capture (7).webm](https://github.com/user-attachments/assets/7d9c787a-9bda-4f24-bd65-30a91b145001)

### Issues Resolved
Resolves #404 
Resolves #407

### Check List

### DIFF
--- a/public/pages/workflow_detail/workflow_inputs/processor_inputs/ml_processor_inputs.tsx
+++ b/public/pages/workflow_detail/workflow_inputs/processor_inputs/ml_processor_inputs.tsx
@@ -45,6 +45,7 @@ import {
   parseModelOutputs,
 } from '../../../../utils';
 import { ConfigFieldList } from '../config_field_list';
+import { OverrideQueryModal } from './modals/override_query_modal';
 
 interface MLProcessorInputsProps {
   uiConfig: WorkflowConfig;
@@ -127,6 +128,7 @@ export function MLProcessorInputs(props: MLProcessorInputsProps) {
     boolean
   >(false);
   const [isPromptModalOpen, setIsPromptModalOpen] = useState<boolean>(false);
+  const [isQueryModalOpen, setIsQueryModalOpen] = useState<boolean>(false);
 
   // model interface state
   const [modelInterface, setModelInterface] = useState<
@@ -281,6 +283,14 @@ export function MLProcessorInputs(props: MLProcessorInputsProps) {
           onClose={() => setIsPromptModalOpen(false)}
         />
       )}
+      {isQueryModalOpen && (
+        <OverrideQueryModal
+          config={props.config}
+          baseConfigPath={props.baseConfigPath}
+          modelInterface={modelInterface}
+          onClose={() => setIsQueryModalOpen(false)}
+        />
+      )}
       <ModelField
         field={modelField}
         fieldPath={modelFieldPath}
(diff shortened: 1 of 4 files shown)


public/pages/workflow_detail/workflow_inputs/processor_inputs/ml_processor_inputs.tsx
```diff
@@ -290,6 +300,23 @@ export function MLProcessorInputs(props: MLProcessorInputsProps) {
...
+                style={{ marginTop: '4px' }}
+              >{`Override query (Optional)`}</EuiText>
+              <EuiSpacer size="s" />
+              <EuiSmallButton
+                style={{ width: '100px' }}
+                fill={false}
+                onClick={() => setIsQueryModalOpen(true)}
+              >
+                Override
+              </EuiSmallButton>
```
saimedhi, line 316: @Tyler, when adding buttons can we also add data-testid. That helps in future when writing tests.

public/pages/workflow_detail/workflow_inputs/processor_inputs/modals/override_query_modal.tsx
```diff
@@ -0,0 +1,263 @@
...
+                </>
+              )}
+            </>
+          </EuiFlexItem>
+        </EuiFlexGroup>
+      </EuiModalBody>
+      <EuiModalFooter>
+        <EuiSmallButton onClick={props.onClose} fill={false} color="primary">
+          Close
+        </EuiSmallButton>
```
saimedhi, line 213: Please add data-testid if possible.

public/pages/workflow_detail/workflow_inputs/processor_inputs/modals/override_query_modal.tsx
```diff
@@ -0,0 +1,263 @@
...
+                                .replace(
+                                  new RegExp(`"${QUERY_IMAGE_PATTERN}"`, 'g'),
+                                  `\$\{query_image\}`
+                                )
+                                .replace(
+                                  new RegExp(`"${MODEL_ID_PATTERN}"`, 'g'),
+                                  `\$\{model_id\}`
+                                )
+                            );
+                          } catch {}
```
saimedhi, line 162: Can we have a failure message within catch?

saimedhi (approved):
Overall, Looks good to me. Thank you Tyler.

public/pages/workflow_detail/workflow_inputs/processor_inputs/ml_processor_inputs.tsx
ohltyler: sure will add

public/pages/workflow_detail/workflow_inputs/processor_inputs/modals/override_query_modal.tsx
ohltyler: Copied this over, and realized we don't need it, since we aren't doing any actions that can throw exceptions. Removed it altogether.

PR #462, opened 2024-11-04
Added drill-down details for workflow-created resources

### LINKED ISSUE #378: [Enhancement] Add drill-down details for workflow-created resources
We currently display the associated resources created from a workflow. But, it would be even more helpful to drill down further into these resources, and view their individual configurations. For example, if a workflow creates an ingest pipeline, I may want to look at the pipeline config in a standalone way. Or, view some created index's generated mappings, or default settings. This can help in overall visibility, and let users decide if they want to override some defaults, or pass around those individual resource configs for deeper testing, etc.

The resources are visible in 2 places. Once on the workflow list page in the right-side flyout, and once under the "Resources" tab in the Tools panel in the workflow detail page:
Screenshot 2024-09-13 at 1 46 18 PM
Screenshot 2024-09-13 at 1 46 30 PM

Ideally, we have some mechanism to click on those resources, or have a button somewhere, to pull up the resource details. Since we know the resource type and resource ID, we can execute the corresponding `GET` call to fetch such details. A few APIs may need to be added on the frontend to support this:
- get ingest pipeline API - https://opensearch.org/docs/latest/ingest-pipelines/get-ingest/#retrieving-information-about-a-specific-pipeline
- get search pipeline API - https://opensearch.org/docs/latest/search-plugins/search-pipelines/retrieving-search-pipeline/
- (maybe) variation of get index API. Depends on what we want to return (settings, mappings, both?)

saimedhi:
### Description

- In the workflow list page, used Expanding rows to display Resource details. (Using Eui Basic table). 
- In the workflow details page (Workflow Editor Page), used flyout to display Resource details. (Using Eui Memory table). 
- Added Node APIs  - getSearchPipeline, getIngestPipeline, getIndex.
- For Index, displaying both settings and mappings in Resouce details. 

Old Look:

[screen-capture (28).webm](https://github.com/user-attachments/assets/c86e4e87-4ba4-4a55-9c0b-2cc7b886240e)

New Look:

[screen-capture (26).webm](https://github.com/user-attachments/assets/16732006-b034-46e3-9cb1-bbd1eaea503f)

### Issues Resolved

closes #378

### Check List

### DIFF
--- a/public/pages/workflow_detail/tools/resources/resource_list_with_flyout.tsx
+++ b/public/pages/workflow_detail/tools/resources/resource_list_with_flyout.tsx
@@ -16,6 +16,8 @@ import {
   EuiTitle,
   EuiIcon,
   EuiText,
+  EuiEmptyPrompt,
+  EuiLoadingSpinner,
 } from '@elastic/eui';
 import {
   Workflow,
@@ -23,9 +25,10 @@ import {
   customStringify,
 } from '../../../../../common';
 import { fetchResourceData } from '../../../../utils';
-import { useAppDispatch } from '../../../../store';
+import { AppState, useAppDispatch } from '../../../../store';
 import { getDataSourceId } from '../../../../utils';
 import { columns } from './columns';
+import { useSelector } from 'react-redux';
 
 interface ResourceListFlyoutProps {
   workflow?: Workflow;
@@ -39,6 +42,7 @@ export function ResourceListWithFlyout(props: ResourceListFlyoutProps) {
   const dispatch = useAppDispatch();
   const dataSourceId = getDataSourceId();
   const [resourceDetails, setResourceDetails] = useState<string | null>(null);
+  const { loading } = useSelector((state: AppState) => state.opensearch);
 
   // Hook to initialize all resources. Reduce to unique IDs, since
   // the backend resources may include the same resource multiple times
@@ -72,10 +76,7 @@ export function ResourceListWithFlyout(props: ResourceListFlyoutProps) {
     try {
       const result = await fetchResourceData(row, dataSourceId!, dispatch);
       setResourceDetails(customStringify(result));
-    } catch (error) {
-      console.error('Failed to fetch resource data:', error);
-      setResourceDetails('Error fetching resource data.');
-    }
+    } catch (error) {}
   };
 
   // Closes the flyout and resets the selected resource data.
@@ -123,21 +124,27 @@ export function ResourceListWithFlyout(props: ResourceListFlyoutProps) {
           </EuiFlyoutHeader>
           <EuiFlyoutBody>
             <EuiFlexGroup direction="column" gutterSize="xs">
-              <EuiFlexItem grow={true} style={{ paddingLeft: '20px' }}>
+              <EuiFlexItem grow={true}>
                 <EuiText size="m">
                   <h4>Resource details</h4>
                 </EuiText>
               </EuiFlexItem>
-
               <EuiFlexItem grow={true}>
-                <EuiCodeBlock
-                  language="json"
-                  fontSize="m"
-                  isCopyable={true}
-                  overflowHeight={650}
-                >
-                  {resourceDetails || 'Loading...'}
-                </EuiCodeBlock>
+                {resourceDetails && !loading ? (
+                  <EuiCodeBlock
+                    language="json"
+                    fontSize="m"
+                    isCopyable={true}
+                    overflowHeight={650}
+                  >
+                    {resourceDetails}
+                  </EuiCodeBlock>
+                ) : (
+                  <EuiEmptyPrompt
+                    icon={<EuiLoadingSpinner size="xl" />}
+                    title={<h2>Loading</h2>}
+                  />
+                )}
               </EuiFlexItem>
             </EuiFlexGroup>
           </EuiFlyoutBody>
--- a/public/pages/workflow_detail/tools/resources/resource_list_with_flyout.tsx
+++ b/public/pages/workflow_detail/tools/resources/resource_list_with_flyout.tsx
@@ -22,7 +22,7 @@ import {
   WorkflowResource,
   customStringify,
 } from '../../../../../common';
-import { fetchResourceData } from '../../../workflows/workflow_list/resource_list';
+import { fetchResourceData } from '../../../../utils';
 import { useAppDispatch } from '../../../../store';
 import { getDataSourceId } from '../../../../utils';
 import { columns } from './columns';
--- a/public/pages/workflow_detail/tools/resources/resource_list_with_flyout.tsx
+++ b/public/pages/workflow_detail/tools/resources/resource_list_with_flyout.tsx
@@ -134,7 +134,7 @@ export function ResourceListWithFlyout(props: ResourceListFlyoutProps) {
                   language="json"
                   fontSize="m"
                   isCopyable={true}
-                  overflowHeight={680}
+                  overflowHeight={650}
                 >
                   {resourceDetails || 'Loading...'}
                 </EuiCodeBlock>
--- a/public/pages/workflow_detail/tools/resources/resource_list_with_flyout.tsx
+++ b/public/pages/workflow_detail/tools/resources/resource_list_with_flyout.tsx
@@ -17,11 +17,15 @@ import {
   EuiIcon,
   EuiText,
 } from '@elastic/eui';
-import { Workflow, WorkflowResource, customStringify } from '../../common';
-import { fetchResourceData } from './resource_list';
-import { useAppDispatch } from '../store';
-import { getDataSourceId } from '../../public/utils';
-import { columns } from '../pages/workflow_detail/tools/resources/columns';
+import {
+  Workflow,
+  WorkflowResource,
+  customStringify,
+} from '../../../../../common';
+import { fetchResourceData } from '../../../workflows/workflow_list/resource_list';
+import { useAppDispatch } from '../../../../store';
+import { getDataSourceId } from '../../../../utils';
+import { columns } from './columns';
 
 interface ResourceListFlyoutProps {
   workflow?: Workflow;
--- a/public/pages/workflow_detail/tools/resources/resource_list_with_flyout.tsx
+++ b/public/pages/workflow_detail/tools/resources/resource_list_with_flyout.tsx
@@ -42,7 +42,7 @@ export function ResourceListWithFlyout(props: ResourceListFlyoutProps) {
   const dispatch = useAppDispatch();
   const dataSourceId = getDataSourceId();
   const [resourceDetails, setResourceDetails] = useState<string | null>(null);
-  const { loading } = useSelector((state: AppState) => state.opensearch);
+  const loading = useSelector((state: AppState) => state.opensearch.loading);
 
   // Hook to initialize all resources. Reduce to unique IDs, since
   // the backend resources may include the same resource multiple times
--- a/public/pages/workflow_detail/tools/resources/resource_list_with_flyout.tsx
+++ b/public/pages/workflow_detail/tools/resources/resource_list_with_flyout.tsx
@@ -24,9 +24,18 @@ import {
   WorkflowResource,
   customStringify,
 } from '../../../../../common';
-import { fetchResourceData } from '../../../../utils';
-import { AppState, useAppDispatch } from '../../../../store';
-import { getDataSourceId } from '../../../../utils';
+import {
+  AppState,
+  useAppDispatch,
+  getIndex,
+  getIngestPipeline,
+  getSearchPipeline,
+} from '../../../../store';
+import {
+  extractIdsByStepType,
+  getDataSourceId,
+  getErrorMessageForStepType,
+} from '../../../../utils';
 import { columns } from './columns';
 import { useSelector } from 'react-redux';
 
@@ -42,7 +51,16 @@ export function ResourceListWithFlyout(props: ResourceListFlyoutProps) {
   const dispatch = useAppDispatch();
   const dataSourceId = getDataSourceId();
   const [resourceDetails, setResourceDetails] = useState<string | null>(null);
-  const loading = useSelector((state: AppState) => state.opensearch.loading);
+  const [rowErrorMessage, setRowErrorMessage] = useState<string | null>(null);
+  const {
+    loading,
+    getIndexErrorMessage,
+    getIngestPipelineErrorMessage,
+    getSearchPipelineErrorMessage,
+    indexDetails,
+    ingestPipelineDetails,
+    searchPipelineDetails,
+  } = useSelector((state: AppState) => state.opensearch);
 
   // Hook to initialize all resources. Reduce to unique IDs, since
   // the backend resources may include the same resource multiple times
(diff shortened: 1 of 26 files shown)


saimedhi:
Integration tests passing.

public/pages/workflows/workflow_list/resource_list.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  useEffect(() => {
+    if (codeBlockData) {
+      const { item, data } = codeBlockData;
+      setItemIdToExpandedRowMap((prevMap) => ({
+        ...prevMap,
+        [item.id]: renderExpandedRow(data),
+      }));
+    }
+  }, [codeBlockData]);
```
ohltyler, line 72: Few enhancements I can see here:
1. I don't think we need this extra custom hook. It seems we can update the `itemIdToExpandedRowMap` with the data directly on line 109, and skip using the additional `codeBlockData` state var entirely
2. We shouldn't need a `codeBlockData` state at all. Instead, we should just dynamically render the code block based on the item ID exposed in the `render` fn for the table column directly on line 160.

public/pages/workflows/workflow_list/resource_list.tsx
```diff
@@ -0,0 +1,207 @@
...
+/**
+ * The searchable list of resources for a particular workflow.
+ */
+export function ResourceList(props: ResourceListProps) {
+  const [allResources, setAllResources] = useState<WorkflowResource[]>([]);
+  const dispatch = useAppDispatch();
+  const dataSourceId = getDataSourceId();
+  const [itemIdToExpandedRowMap, setItemIdToExpandedRowMap] = useState<{
+    [key: string]: React.ReactNode;
+  }>({});
```
ohltyler, line 44: This is inefficient for persisting the data - we should separate the data from the rendered components to save space and to simplify the data model design. I'd suggest having a map mapping ID -> string, instead of ID -> ReactNode. The `render` fn on line 160 can then take the exposed item ID, and render the ReactNode (in this case, the `EuiCodeBlock`) using the map's string value directly.

This way, we only persist the minimal state required, and let the components handle the dynamic rendering of the data.

public/pages/workflows/workflow_list/resource_list.tsx
```diff
@@ -0,0 +1,207 @@
...
+      })
+    ).unwrap();
+  } else if (item.stepType.includes('search_pipeline')) {
+    return await dispatch(
+      getSearchPipeline({
+        pipelineId: item.id,
+        dataSourceId,
+      })
+    ).unwrap();
+  }
```
ohltyler, line 205: Please use the `WORKFLOW_STEP_TYPE` constants in common/interfaces.ts to determine which dispatch calls to execute.

public/pages/workflow_detail/tools/resources/resource_list_with_flyout.tsx
```diff
@@ -0,0 +1,148 @@
...
+  EuiTitle,
+  EuiIcon,
+  EuiText,
+} from '@elastic/eui';
+import {
+  Workflow,
+  WorkflowResource,
+  customStringify,
+} from '../../../../../common';
+import { fetchResourceData } from '../../../workflows/workflow_list/resource_list';
```
ohltyler, line 25: If you'd like to reuse this fn, let's lift it up to a shared utils file somewhere so then it is logically consumed by the downstream components.

server/routes/opensearch_routes_service.ts
```diff
@@ -429,4 +526,59 @@ export class OpenSearchRoutesService {
...
+        this.dataSourceEnabled,
+        req,
+        data_source_id,
+        this.client
+      );
+
+      const response = await callWithRequest('transport.request', {
+        method: 'GET',
+        path: SEARCH_PIPELINE_ROUTE + '/' + pipeline_id,
+      });
```
ohltyler, line 577: Is there not a node js client path for getting search pipelines, like for the other core apis? (e.g., `ingest.getPipeline`)?

If not, I'd suggest to add this path in `server/cluster` somewhere. Note we have the defined paths to execute different plugin APIs for Flow Framework and ML commons there. Maybe, would need a new "core" or "search pipelines" plugin to logically persist them there.

public/store/reducers/opensearch_reducer.ts
ohltyler, line 1: 2 suggestions here:
1. Please add the pending/fulfilled/rejected states in the `opensearchSlice` defined on line 261. This is where the reducer really helps and persists the client-side state for us. As the calls are being executed, the opensearch state is updated accordingly (loading state, the persisted indices, etc.)
2. The state for this reducer can be expanded to persist ingest & search pipelines state; then we don't have to do any state management of these resources in the components themselves. You can follow the pattern we have for persisting the indices. Then, no `itemIDToExpandedRowMap` vars etc. that you have in `ResourceList` will be needed at all; instead, just pull the data persisted in the reducer state with `useSelector`.

ohltyler: I'd recommend looking at examples of how we fetch and display other reducer state, such as indices, or flow framework workflows. Typically, you can just send out one dispatch call to fetch what's needed, and then use `useSelector` to pull the formatted data. It will be automatically updated when the APIs return their responses. You may reference https://redux.js.org/tutorials/fundamentals/part-1-overview for more details.

public/pages/workflows/workflow_list/resource_list.tsx
ohltyler: Update: above may be irrelevant after addressing https://github.com/opensearch-project/dashboards-flow-framework/pull/462#discussion_r1828266467

public/pages/workflows/workflow_list/resource_list.tsx
```diff
@@ -0,0 +1,207 @@
...
+  // Handles table pagination and sorting.
+  const onTableChange = ({
+    page = { index: pageIndex, size: pageSize },
+    sort = { field: 'id', direction: 'asc' },
+  }: any) => {
+    setPageIndex(page.index);
+    setPageSize(page.size);
+    setSortField(sort.field as keyof WorkflowResource);
+    setSortDirection(sort.direction);
+  };
```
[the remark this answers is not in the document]
saimedhi: - @ohltyler, Existing  resource_list uses EuiInMemoryTable. So, pagination and sorting are handled within it. 
- Now to add expanding rows to display resource details, I am not able to do with EuiInMemoryTable. So, using EuiBasicTable. For EuiBasicTable we need to handle pagination and sorting.

public/pages/workflows/workflow_list/resource_list.tsx
```diff
@@ -0,0 +1,207 @@
...
+      ? -1 * multiplier
+      : 0;
+  });
+
+  const pagination = {
+    pageIndex,
+    pageSize,
+    totalItemCount: allResources.length,
+    pageSizeOptions: [10, 25, 50],
+  };
```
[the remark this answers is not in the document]
saimedhi: @ohltyler, Existing resource_list uses EuiInMemoryTable. So, pagination and sorting are handled within it.
Now to add expanding rows to display resource details, I am not able to do with EuiInMemoryTable. So, using EuiBasicTable. For EuiBasicTable we need to handle pagination and sorting.

server/routes/opensearch_routes_service.ts
saimedhi: @ohltyler, I initially tried by adding in server/cluster. But after referring AD frontend plugin [here](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/d8f738076e265dc858800573529469ca6d13ce19/server/routes/opensearch.ts#L176) ,  using `transport.request `felt right. Instead of adding other file.

public/pages/workflows/workflow_list/resource_list.tsx
ohltyler: Update 2: I see this map is required for using [expanded rows](https://oui.opensearch.org/1.6/#/tabular-content/tables#expanding-rows), so I'm fine with using that. But still prefer to fetch all of the data from redux store, and create/update the map based on the data from that store instead of the current mechanism of fetching and parsing the API response directly.

server/routes/opensearch_routes_service.ts
ohltyler: Ok, my concern is that this is objectively weaker and we don't have as much type safety or guardrails around the API than if we explicitly define, AD should not have done it that way. Not a blocker though. Also, please double check the JS client as it seems this should be available out of the box. Not sure if search pipelines APIs got missed?

saimedhi: Will followup on this

public/pages/workflow_detail/tools/resources/resource_list_with_flyout.tsx
```diff
@@ -0,0 +1,148 @@
...
+  // Opens the flyout and fetches resource details for the selected row.
+  const openFlyout = async (row: WorkflowResource) => {
+    setSelectedRowData(row);
+    setIsFlyoutVisible(true);
+    try {
+      const result = await fetchResourceData(row, dataSourceId!, dispatch);
+      setResourceDetails(customStringify(result));
+    } catch (error) {
+      console.error('Failed to fetch resource data:', error);
+      setResourceDetails('Error fetching resource data.');
```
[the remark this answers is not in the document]
kamingleung: Are there reasons it will failed to fetch resource details? Permissions?

public/pages/workflow_detail/tools/resources/resource_list_with_flyout.tsx
```diff
@@ -0,0 +1,148 @@
...
+              </EuiFlexItem>
+
+              <EuiFlexItem grow={true}>
+                <EuiCodeBlock
+                  language="json"
+                  fontSize="m"
+                  isCopyable={true}
+                  overflowHeight={650}
+                >
+                  {resourceDetails || 'Loading...'}
```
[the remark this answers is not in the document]
kamingleung: With the [OuiEmptyPrompt](https://oui.opensearch.org/1.17/#/display/empty-prompt#loading-and-error-prompts) component, we can show a loading state like this:
image
We use [OuiLoadingSpinner](https://oui.opensearch.org/1.17/#/display/loading#spinner) for the spinner icon.

kamingleung:
In the resource details modal (opened from the workflow detail page/visual editor), The "Resource details" header should align with all of the content:
image

saimedhi:
Below screenshots include resource details displayed in workflow details page(editor page - flyout used here) and workflow list page (expanding rows used here). And also included screenshots of error and loading scenarios. 

Screenshot 2024-11-05 at 8 42 57 AM
Screenshot 2024-11-05 at 8 43 40 AM
Screenshot 2024-11-05 at 8 48 31 AM
Screenshot 2024-11-05 at 8 55 33 AM
Screenshot 2024-11-05 at 8 44 19 AM
Screenshot 2024-11-05 at 8 44 36 AM
Screenshot 2024-11-05 at 8 46 47 AM

kamingleung:
@saimedhi Thanks for your updates! Is the error message hard coded? Are we able to display different messages for different failures?
image

saimedhi:
> @saimedhi Thanks for your updates! Is the error message hard coded? Are we able to display different messages for different failures? image


Displaying failure messages, Example:
Screenshot 2024-11-05 at 9 47 19 AM

public/pages/workflows/workflow_list/resource_list.tsx
saimedhi: Removed.

public/pages/workflows/workflow_list/resource_list.tsx
```diff
@@ -0,0 +1,207 @@
...
+  );
+
+  // Expands or collapses the details for a resource item.
+  const toggleDetails = async (item: WorkflowResource) => {
+    const updatedItemIdToExpandedRowMap = { ...itemIdToExpandedRowMap };
+
+    if (updatedItemIdToExpandedRowMap[item.id]) {
+      delete updatedItemIdToExpandedRowMap[item.id];
+      setItemIdToExpandedRowMap(updatedItemIdToExpandedRowMap);
+    } else {
```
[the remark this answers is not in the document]
saimedhi: Needed for expanding rows

public/pages/workflows/workflow_list/resource_list.tsx
```diff
@@ -0,0 +1,188 @@
...
+  const renderExpandedRow = useCallback(
+    (data: any) => (
+      <EuiFlexGroup direction="column" gutterSize="xs">
+        <EuiFlexItem grow={true} style={{ paddingLeft: '20px' }}>
+          <EuiText size="m">
+            <h4>Resource details</h4>
+          </EuiText>
+        </EuiFlexItem>
+        <EuiFlexItem grow={true} style={{ paddingLeft: '20px' }}>
+          {!data.startsWith('error:') && !loading ? (
```
ohltyler, line 73: This is not a clean way to propagate errors. Additionally, the input `data` is not typesafe / defined as a string, so we cannot assume we can leverage `startsWith()`. 

This is another situation where this state can be much more cleanly passed via the redux store, the same way as the `loading` var is. In fact, you already have the `errorMessage` var populated. Let's leverage this, and then just dynamically display the `EuiEmptyPrompt` with the error message if `errorMessage` is defined.

public/pages/workflows/workflow_list/resource_list.tsx
ohltyler: The results from search index are not persisted because they are transient/runtime values, only needed at the particular time of execution. Redux store is for persisting/caching backend resources / data that is used across the UI, such that it is not needed to be fetched constantly. Note how many of the resource fetching of current backend resources (workflows, indices, ML models etc.) are fetched in the top-level components once to populate the redux store. Then, the downstream child components simply fetch from that store, without having to rerun the APIs and increase traffic to the cluster.

Ingest and search pipelines are persistent resources, and should be fetched in the same way. Then, the tables that leverage them don't need to make extra API calls. Additionally, the resources are then persisted in a standard, defined way. Note how we have defined interfaces for Workflows, Models, etc. that are formatted and saved in the redux store. Then, when consumed by other components, it is entirely typesafe and the data model is entirely known.

Can you explain what you mean by `Fetching all Resources Details and storing in redux Store doesn't seem optimal`? I'd consider this a blocker still for this PR, since it's directly introducing tech debt by the current implementation.

server/routes/opensearch_routes_service.ts
ohltyler: IMHO it's less effort to make the change then follow up with tracking issues etc. 😄  But please do one of them before resolving the conversation so it is tracked accordingly.

saimedhi: This is added [here](https://github.com/opensearch-project/dashboards-flow-framework/pull/462/commits/b5c9818164e9034b4bd2ac95378659d4a6a8ae85)

public/pages/workflows/workflow_list/resource_list.tsx
saimedhi: - Tyler, As discussed I am working on **Fetching all resource details** (meaning details of all resources)present in a workflow and having them in store while rendering resourceList table . 
- Then from store will fetch details of any resource instead of calling APIs directly. 



Few Insights as per our discussion :

1. Advantages - 

- Uniformity across plugin to fetch to store and use it from there.
- When user looks at same resource details multiple times we are not doing API calls several times.  

2. Disadvantages -

- Though the user is not looking for resource details in the drill down and just opened resource list we will do API calls to get all the resources Details in that workflow into Store.

saimedhi:
@ohltyler  reimplemented API calls, Error Message handling as suggested. 

Screenshot 2024-11-06 at 1 42 53 PM
Screenshot 2024-11-06 at 1 45 31 PM
Screenshot 2024-11-06 at 1 44 37 PM
Screenshot 2024-11-06 at 1 46 17 PM

public/pages/workflows/workflow_list/resource_list.tsx
```diff
@@ -0,0 +1,240 @@
...
+      try {
+        dispatch(
+          getSearchPipeline({ pipelineId: searchPipelineIds, dataSourceId })
+        );
+      } catch {}
+    }
+  }, [allResources]);
+
+  // Renders the expanded row to show resource details in a code block.
+  const renderExpandedRow = useCallback(
```
ohltyler, line 111: Why use `useCallback` here, and why is it only updated if `loading` is updated? I think we can remove.

saimedhi: removed useCallback. Initially added this to  re-render expanding row content in case of loading changes from True to false.

ohltyler (approved):
LGTM! Please be mindful of force pushing during reviews, as it makes it much more difficult to track the changes made over subsequent commits.

PR #481, opened 2024-11-15
Improve consistency of JSONPath parsing & execution across frontend/backend

### LINKED ISSUE #478: Improve handling edge cases regarding JSONPath / dot notation discrepancies
There are some remaining discrepancies in the ML processor transforms regarding dot notation and JSONPath, that can yield different results when rendering examples on the frontend, vs. the execution on the backend. As such, this can result in successful transformations when testing on the frontend, but failures when executing ingest/search. Tracking known discrepancies below:

- [x] ML search request/response transforms only use JSONPath, and the [backend JSONPath package](https://github.com/json-path/JsonPath) does not support dot notation for arrays. To have feature parity, we should automatically convert any dot notation with arrays (e.g., `arrays.0`) into dot notation with brackets (e.g., `arrays[0]`)., or see if the functionality to automatically generate the keys, can take in some config to automatically convert to brackets by default, or see if there is some external package that can handle all of this dot -> bracket transformation for us altogether.  
- [x] discrepancies with arrays for single values. Currently, the default JSONPath [query command](https://www.npmjs.com/package/jsonpath#jpqueryobj-pathexpression-count) we execute returns an array by default, even if for a discrete/single/definite value. The backend package has a config field called `ALWAYS_RETURN_LIST`, set to `false`. We should see if there is a similar config on the frontend side, and/or if we can parse out whether a path is "definite" or not, and do post-processing.
- [x] enforce output transforms are always using JSONPath as required per the backend.
- [x] improve tooltip wording to indicate the above discrepancies / limitations
- [x] keys with dashes in them fail using the nodejs library - for example, given:
```
{ 
  "field": {
    "sub-field": "value" 
  }
}
```
and some jsonpath, like `$.field.sub-field`, will fail to fetch "value". One workaround is to wrap it in brackets with quotes - for example, `$.field["sub-field"]` will succeed.

ohltyler:
### Description

This PR improves the path parsing and JSONPath handling when configuring advanced transforms in the ML processors. Added detailed comments in the source code for readability.

- improves `generateTransform` to handle more edge cases and be consistent without the backend handles path parsing for the different scenarios (input vs. output, ingest vs. search request vs. search response).
- adds new`sanitizeJSONPath()` fn used in 2 places: 1/ when auto-populating keys for the query or document JSONs (ensures they will run consistently on frontend & backend), and 2/ when converting to a template (to ensure it runs properly on the backend).
- adds new fn `isIndefiniteJsonPath()` to help make the return values of a JSONPath execution consistent. Note: this is not a full-featured fn. The frontend libraries do not have any configuration similar to `ALWAYS_RETURN_LIST`, nor a mechanism to determine whether a path is definite or indefinite. The purpose of this function is to broadly cover the majority of use cases to mimic the implementation as close as possible.

More details on the fixes this covers in the related issue #478 .

### Issues Resolved

Resolves #478 

### Check List

### DIFF
--- a/public/utils/utils.ts
+++ b/public/utils/utils.ts
@@ -18,6 +18,7 @@ import {
   PROCESSOR_CONTEXT,
   SimulateIngestPipelineDoc,
   SimulateIngestPipelineResponse,
+  TRANSFORM_CONTEXT,
   WORKFLOW_RESOURCE_TYPE,
   WORKFLOW_STEP_TYPE,
   Workflow,
@@ -186,7 +187,8 @@ export function unwrapTransformedDocs(
 export function generateTransform(
   input: {} | [],
   map: MapFormValue,
-  context: PROCESSOR_CONTEXT
+  context: PROCESSOR_CONTEXT,
+  transformContext: TRANSFORM_CONTEXT
 ): {} {
   let output = {};
   map.forEach((mapEntry) => {
@@ -195,7 +197,8 @@ export function generateTransform(
         mapEntry,
         input,
         mapEntry.value,
-        context
+        context,
+        transformContext
       );
       output = {
         ...output,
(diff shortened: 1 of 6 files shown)


public/utils/utils.ts
```diff
@@ -223,19 +233,60 @@ export function generateArrayTransform(input: [], map: MapFormValue): {}[] {
...
-  // and the input is the entire result with nothing else to parse out.
-  // get() does not cover this case, so we override manually.
-  return path === '.'
-    ? input
-    : mapEntry.value.startsWith(JSONPATH_ROOT_SELECTOR)
-    ? // JSONPath transform
-      jsonpath.query(input, path)
-    : // Standard dot notation
-      get(input, path);
+  // Rregular dot notation can only be executed if 1/ the JSONPath selector is not explicitly defined,
```
saimedhi, line 239: nit : Regular

saimedhi (approved):

PR #503, opened 2024-11-27
Remove legacy preview modals; add field-level input validation

ohltyler:
### Description

This PR completes the initial UX design changes to the new input & output transforms for the ML processors.
- removes the legacy unused modals and unused code in MLProcessorInputs
- adds field-level validation checks when configuring input transforms. Validation can be expanded in the future, we may show checks per-field in the base form instead of just the modal, etc. but that is a separate effort & designs have not been finalized
- bug fix of caching issue in the modals when multiple modals may be triggered from one set of inputs/outputs, that could sometimes render outdated/invalid state in the modals when reopening. Fixes this by only persisting the mapping index, instead of a single 'isOpen' var

Demo video showing the field-level validation (both when a field is found in the interface, and hiding when no interface found), and showing that multiple modals don't trigger any outdated state. 

[screen-capture (11).webm](https://github.com/user-attachments/assets/97e9aa15-036e-4313-80c7-07adbca6d3f2)

### Check List


saimedhi (approved):
LGTM. Will also check by running this feature branch locally once this PR is merged into feature branch.

PR #530, opened 2024-12-12
Fix MDS bug

saimedhi:
### Description

- If mds is enabled, Added code to wait until selected data source is ready before doing dispatch calls 
- This is done to ensure workflow page(initial page) is properly loaded when mds is enabled. 
- Failing tests unrelated to this PR. 

### Check List

### DIFF
--- a/public/pages/workflows/workflows.tsx
+++ b/public/pages/workflows/workflows.tsx
@@ -153,7 +153,7 @@ export function Workflows(props: WorkflowsProps) {
         })
       );
     }
-  }, []);
+  }, [dataSourceId, dataSourceEnabled]);
 
   useEffect(() => {
     const { history, location } = props;
@@ -176,7 +176,7 @@ export function Workflows(props: WorkflowsProps) {
         })
       );
     }
-  }, [dataSourceId, setDataSourceId]);
+  }, [dataSourceId, setDataSourceId, dataSourceEnabled]);
 
   const handleDataSourceChange = ([event]: DataSourceOption[]) => {
     const dataSourceEventId = event?.id;
--- a/public/pages/workflows/workflows.tsx
+++ b/public/pages/workflows/workflows.tsx
@@ -102,6 +102,9 @@ export function Workflows(props: WorkflowsProps) {
     escape(tabFromUrl) as WORKFLOWS_TAB
   );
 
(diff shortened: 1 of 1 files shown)


public/pages/workflows/workflows.tsx
```diff
@@ -116,12 +116,15 @@ export function Workflows(props: WorkflowsProps) {
...
   useEffect(() => {
     if (selectedTabId === WORKFLOWS_TAB.MANAGE) {
-      dispatch(
-        searchWorkflows({
-          apiBody: FETCH_ALL_QUERY,
-          dataSourceId: dataSourceId,
-        })
-      );
+      // wait until selected data source is ready before doing dispatch calls if mds is enabled
+      if (!dataSourceEnabled || (dataSourceId && dataSourceId !== '')) {
```
ohltyler, line 120: nit: lets move this boolean into a constant for readability and since it's used in other places.

public/pages/workflows/workflows.tsx
ohltyler, line 1: is this needed at top-level of workflow details page as well?

public/pages/workflows/workflows.tsx
ohltyler: This hook will only be triggered on changes to `selectedTabId`, hence if the data source ID comes in async later, this won't be re-run. Suggest to add these dependent vars into the list of listening vars alongside `selectedTabId` on line 129.

public/pages/workflows/workflows.tsx
saimedhi: I don't think workflow details page will need this. workflow details page will always have datasourceId passed to url when MDS enabled.

public/pages/workflows/workflows.tsx
saimedhi: Tyler, this case is handled here https://github.com/opensearch-project/dashboards-flow-framework/pull/530/files#diff-a3a1d6b511eaf420acbca3b67675dae82fd466bac8d2a8cf5f3e2aee7392424fR158

ellazhu1022 (approved):

huyaboo (approved):

ohltyler (approved):

saimedhi:
Tested again. MDS Working fine in all cases.

Issue #550, opened 2024-12-27
[ENHANCEMENT] hide ProcessorList component for SEARCH_REQUEST when version < 2.19.0


kamahuan:
Currently, the ProcessorsList component is displayed for all contexts regardless of version. For better version compatibility, we need to hide the entire processor section in the WorkflowInputs component when:
- Context is SEARCH_REQUEST
- Version is less than 2.19.0

This change should be implemented in the `WorkflowInputs` component, not in the `ProcessorsList` component.
Reference: https://github.com/opensearch-project/dashboards-flow-framework/pull/537#discussion_r1893275265

ohltyler:
Closed as part of #605

PR #600, opened 2025-02-06
hide search query section when version is less than 2.19

kamahuan:
### Description

Upon discussion with the UX team, we need to hide the whole transform query section if backend version is less than 2.19. The issue states to hide the processor list section. We need to hide the whole query section instead. 

### Issues Resolved

[550](https://github.com/opensearch-project/dashboards-flow-framework/issues/550)

### Check List

### DIFF
--- a/public/pages/workflow_detail/workflow_inputs/search_inputs/enrich_search_request.tsx
+++ b/public/pages/workflow_detail/workflow_inputs/search_inputs/enrich_search_request.tsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import React from 'react';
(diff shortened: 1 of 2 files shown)


ohltyler:
How about hiding the rendering of the component altogether if the version is <2.19? Then, you don't need to add any logic to fetch the data source ID. The parent component `SearchInputs` already has the data source ID.

Additionally, suggest to use `semver` for handling the version checks, like what you've done [here](https://github.com/opensearch-project/dashboards-flow-framework/blob/main/public/pages/workflow_detail/workflow_inputs/processors_list.tsx#L142) for example prior.

ohltyler:
Closing in favor of #605 .

PR #606, opened 2025-02-07
fix error that local cluster cannot get version

kamahuan:
### Description

Currently in the new work flow page, when user click local data source, it shows a loading state and does not render any presets. This PR fix this issue by getting version from the datasource and render the presets based on the data source version.

### Issues Resolved

None

#### for version 2.19
Xnip2025-02-07_00-07-07
#### for version 2.17
Xnip2025-02-07_00-16-27

### Check List

### DIFF
--- a/public/pages/workflow_detail/workflow_inputs/processors_list.tsx
+++ b/public/pages/workflow_detail/workflow_inputs/processors_list.tsx
@@ -119,7 +119,7 @@ export function ProcessorsList(props: ProcessorsListProps) {
       return;
     }
 
-    if (dataSourceId) {
+    if (dataSourceId !== undefined) {
       getEffectiveVersion(dataSourceId)
         .then((ver) => {
           setVersion(ver);
--- a/public/pages/workflows/new_workflow/new_workflow.tsx
+++ b/public/pages/workflows/new_workflow/new_workflow.tsx
@@ -116,6 +116,7 @@ export function NewWorkflow(props: NewWorkflowProps) {
   // 1. fetch the workflow presets persisted on server-side
   // 2. fetch the ML models and connectors. these may be used in quick-create views when selecting a preset,
   //    so we optimize by fetching once at the top-level here.
+  // 3. fetch local cluster version if applicable
   useEffect(() => {
     dispatch(getWorkflowPresets());
     if (isDataSourceReady(dataSourceId)) {
@@ -124,6 +125,7 @@ export function NewWorkflow(props: NewWorkflowProps) {
         searchConnectors({ apiBody: FETCH_ALL_QUERY_LARGE, dataSourceId })
       );
     }
+    // if use local cluster
     if (dataSourceId === '') {
       dispatch(getLocalClusterVersion());
     }
--- a/public/pages/workflows/new_workflow/new_workflow.tsx
+++ b/public/pages/workflows/new_workflow/new_workflow.tsx
@@ -39,6 +39,7 @@ import {
   MIN_SUPPORTED_VERSION,
   MINIMUM_FULL_SUPPORTED_VERSION,
 } from '../../../../common/constants';
+import { getLocalClusterVersion } from '../../../store/reducers/opensearch_reducer';
 
 interface NewWorkflowProps {}
 
@@ -52,7 +53,7 @@ const filterPresetsByVersion = async (
     return workflows;
   }
 
-  if (!dataSourceId) {
+  if (dataSourceId === undefined) {
     return [];
   }
 
@@ -92,14 +93,18 @@ export function NewWorkflow(props: NewWorkflowProps) {
   const dataSourceId = getDataSourceId();
   const dataSourceEnabled = getDataSourceEnabled().enabled;
   // workflows state
-  const { presetWorkflows, loading } = useSelector(
+  const { presetWorkflows, loading: presetsLoading } = useSelector(
     (state: AppState) => state.presets
   );
+  const { loading: opensearchLoading, localClusterVersion } = useSelector(
+    (state: AppState) => state.opensearch
+  );
+  const isLoading = presetsLoading || opensearchLoading;
+
   const [allWorkflows, setAllWorkflows] = useState<WorkflowTemplate[]>([]);
   const [filteredWorkflows, setFilteredWorkflows] = useState<
     WorkflowTemplate[]
   >([]);
-  const [isVersionLoading, setIsVersionLoading] = useState(false);
 
   // search bar state
   const [searchQuery, setSearchQuery] = useState<string>('');
@@ -118,6 +123,9 @@ export function NewWorkflow(props: NewWorkflowProps) {
       dispatch(
         searchConnectors({ apiBody: FETCH_ALL_QUERY_LARGE, dataSourceId })
       );
+    }
+    if (dataSourceId === '') {
+      dispatch(getLocalClusterVersion());
     }
   }, [dataSourceId, dataSourceEnabled]);
 
@@ -140,18 +148,14 @@ export function NewWorkflow(props: NewWorkflowProps) {
         );
         setAllWorkflows(enrichedWorkflows);
         setFilteredWorkflows(enrichedWorkflows);
-        setIsVersionLoading(false);
         return;
       }
 
-      if (!dataSourceId) {
+      if (dataSourceId === undefined) {
         setAllWorkflows([]);
         setFilteredWorkflows([]);
-        setIsVersionLoading(true);
         return;
       }
-
-      setIsVersionLoading(true);
 
       const version = await getEffectiveVersion(dataSourceId);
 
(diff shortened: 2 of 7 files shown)


server/routes/opensearch_routes_service.ts
```diff
@@ -43,6 +43,17 @@ export function registerOpenSearchRoutes(
...
+    {
+      path: `${BASE_NODE_API_PATH}/version`,
+      validate: {
+        params: schema.object({
+          pattern: schema.string(),
+        }),
+      },
+    },
+    opensearchRoutesService.catIndices
+  );
```
ohltyler, line 56: This route is not being used. If adding a server-side call, let's properly integrate with the redux store for consistency.

server/routes/opensearch_routes_service.ts
```diff
@@ -406,6 +417,39 @@ export class OpenSearchRoutesService {
...
+      const cleanedIndices = response.map((index: any) => ({
+        name: index.index,
+        health: index.health,
+      })) as Index[];
+
+      return res.ok({ body: cleanedIndices });
+    } catch (err: any) {
+      return generateCustomError(res, err);
+    }
+  };
```
ohltyler, line 451: I'm confused about this call - why is cat indices needed?

public/pages/workflows/new_workflow/new_workflow.tsx
```diff
@@ -71,7 +78,7 @@ const filterPresetsByVersion = async (
     return workflows;
   }
 
-  if (!dataSourceId) {
+  if (dataSourceId === undefined) {
```
ohltyler, line 81: I remember some discrepancy on how empty string vs. undefined is handled for data sources. @saimedhi please confirm if this change makes sense.

public/pages/workflows/new_workflow/new_workflow.tsx
```diff
@@ -48,6 +48,13 @@ export const getEffectiveVersion = async (
       throw new Error('Data source is required');
     }
 
+    if (dataSourceId === '') {
+      const response = await getCore().http.post('/api/console/proxy', {
+        query: { path: '/', method: 'GET', dataSourceId: '' },
+      });
+      return response.version.number;
+    }
+
```
ohltyler, line 57: Let's integrate this properly into redux and our server-side route services.

server/routes/opensearch_routes_service.ts
kamahuan: Yes this part is not used. Has removed the unused code.

server/routes/opensearch_routes_service.ts
kamahuan: removed this part.

public/pages/workflows/new_workflow/new_workflow.tsx
kamahuan: For the scenario of local cluster, data source id would be an empty string ''. `!dataSourceId` would be true here so it will return empty presets for local cluster.

This is not the behavior we want so I changed the condition to only return empty if data source id is undefined.

public/pages/workflows/new_workflow/new_workflow.tsx
kamahuan: Done. Add check local cluster version check into `route.service` file and redux into `presetRedux` file

public/pages/workflows/new_workflow/new_workflow.tsx
```diff
@@ -48,6 +50,17 @@ export const getEffectiveVersion = async (
       throw new Error('Data source is required');
     }
 
+    // if (dataSourceId === '') {
+    //   const response = await getCore().http.post('/api/console/proxy', {
+    //     query: { path: '/', method: 'GET', dataSourceId: '' },
+    //   });
+    //   return response.version.number;
+    // }
```
saimedhi, line 58: @kamahuan , could you remove this commented lines.

public/pages/workflows/new_workflow/new_workflow.tsx
```diff
@@ -110,14 +123,14 @@ export function NewWorkflow(props: NewWorkflowProps) {
...
-  const { presetWorkflows, loading } = useSelector(
+  const { presetWorkflows, loading, versionLoading } = useSelector(
     (state: AppState) => state.presets
   );
   const [allWorkflows, setAllWorkflows] = useState<WorkflowTemplate[]>([]);
   const [filteredWorkflows, setFilteredWorkflows] = useState<
     WorkflowTemplate[]
   >([]);
-  const [isVersionLoading, setIsVersionLoading] = useState(false);
+  // const [isVersionLoading, setIsVersionLoading] = useState(false);
```
saimedhi, line 133: Please remove commented lines.

saimedhi (commented):
@kamahuan, please clean up the commented code.  Thank you for the fixes.

public/pages/workflows/new_workflow/new_workflow.tsx
saimedhi: **Thank you for fixing the code added in "filter presets and processors based on backend version" PR.** 

Lets connect and test all the possible scenarios before merging this PR.

public/store/reducers/presets_reducer.ts
ohltyler, line 1: In general, I'm ok with the change, but it should go to a more logical reducer - suggest to move to [opensearch_reducer](https://github.com/opensearch-project/dashboards-flow-framework/blob/main/public/store/reducers/opensearch_reducer.ts). Also, after moving, suggest to reuse the existing `loading` state var there; a new one shouldn't be needed.

ohltyler (changes requested):

public/pages/workflows/new_workflow/new_workflow.tsx
kamahuan: done

public/pages/workflow_detail/workflow_inputs/processors_list.tsx
```diff
@@ -110,7 +110,7 @@ export function ProcessorsList(props: ProcessorsListProps) {
       return;
     }
 
-    if (dataSourceId) {
```
jackiehanyang, line 113: can we check the mds feature flag to determine if mds is enabled instead of checking if dataSourceId is present? Checking the feature flag should be the best practice and less error-prone

jackiehanyang: something like this - https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/pages/DetectorsList/containers/List/List.tsx#L147

public/pages/workflows/new_workflow/new_workflow.tsx
jackiehanyang: can we check the mds feature flag to determine if mds is enabled instead of checking if dataSourceId is present? Checking the feature flag should be the best practice and less error-prone

something like this - something like this - https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/pages/DetectorsList/containers/List/List.tsx#L147

ohltyler (changes requested):

public/pages/workflows/new_workflow/new_workflow.tsx
```diff
@@ -117,9 +117,11 @@ export function NewWorkflow(props: NewWorkflowProps) {
   const dataSourceId = getDataSourceId();
   const dataSourceEnabled = getDataSourceEnabled().enabled;
   // workflows state
-  const { presetWorkflows, loading, versionLoading } = useSelector(
-    (state: AppState) => state.presets
+  const { presetWorkflows } = useSelector((state: AppState) => state.presets);
+  const { loading, localClusterVersion } = useSelector(
+    (state: AppState) => state.opensearch
   );
```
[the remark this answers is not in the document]
kamahuan: Adding loading state from the presetReducer back. See line 120. Combining 2 states together, see line: 126

public/store/reducers/presets_reducer.ts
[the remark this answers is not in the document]
kamahuan: Add loading state back to the reducer. See line 12, 43, 50, 54

ohltyler (commented):
LGTM - please resolve the merge conflicts, and work with Sai to ensure everything is tested and working as expected. Thanks!

public/pages/workflows/new_workflow/new_workflow.tsx
jackiehanyang: > Line 76 and 77 checks if MDS is disabled

Don't use `if(dataSourceId === undefined)` to check if MDS is disabled, read the feature flag from config file and check the feature flag to determine if MDS is enabled. Something like this `if(MDSEnabled)` - https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/public/services.ts#L65

ohltyler (approved):

public/pages/workflows/new_workflow/new_workflow.tsx
```diff
@@ -119,6 +124,9 @@ export function NewWorkflow(props: NewWorkflowProps) {
         searchConnectors({ apiBody: FETCH_ALL_QUERY_LARGE, dataSourceId })
       );
     }
+    if (dataSourceId === '') {
+      dispatch(getLocalClusterVersion());
+    }
   }, [dataSourceId, dataSourceEnabled]);
```
ohltyler, line 130: Shouldn't we only run this if data source is enabled, so we don't do unnecessary fetching or error handling? Additionally, please update the comment on lines 115-118 to include this change.

ohltyler (changes requested):

public/pages/workflows/new_workflow/new_workflow.tsx
saimedhi: - Tyler ,  This case only runs when local cluster is used. 

Local cluster - dataSourceId is ""
No dataSource case - datSourceId is undefined.

ohltyler (approved):

PR #537, opened 2024-12-15
filter presets and processors based on backend version

kamahuan:
Filter presets and processors based on backend version

### Description

This PR consists of 3 parts:
1. Filter presets based on backend version. If backend version is 2.17 or 2.18, show presets which not use ML processors. Keep all the presets for version 2.19.
2. Remove default ML processor for the enrich data section for both ingest and search pipeline if version is 2.17 or 2.18.
3. Add ML processor and remove neural processors for version 2.19. Remove ML processors and keep neural processors for version 2.17 and 2.18.

If no datasource is configured during the first render, the create work flow page will have a loading state.

page-loading

### Issues Resolved
NONE

### Check List

- [ ] Commits are signed per the DCO using `--signoff`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #608, opened 2025-02-07
adding prompt when datasource is not compatible

saimedhi:
### Description

- This case is applicable to when datasource is enabled. 
- And when the current using datasources are not compatible with opensearch flow. Then datasourceId will not be present in url. 

- In the above mentioned case, displaying prompt message for users saying datasources are incompatible and add more compatible datasources. 

### Issues Resolved

https://github.com/opensearch-project/dashboards-flow-framework/pull/537#issuecomment-2635518053

### Check List

### DIFF
--- a/public/pages/workflows/new_workflow/new_workflow.tsx
+++ b/public/pages/workflows/new_workflow/new_workflow.tsx
@@ -11,6 +11,8 @@ import {
   EuiFlexGroup,
   EuiCompressedFieldSearch,
   EuiLoadingSpinner,
+  EuiEmptyPrompt,
+  EuiButton,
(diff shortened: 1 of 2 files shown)


saimedhi:
Screenshot 2025-02-07 at 1 16 21 PM

ohltyler (changes requested):
Suggest 2 things:
1. Can we have the messaging cover the entire page, instead of nested in the `New workflow` tab? That prevents any other changes of side effects from users clicking on other things, such as import
2. The wording "incompatible data source" does not make sense, since there is no data source selected. Please confirm with UX on the wording.

PR #616, opened 2025-02-12
Added popover to display links to suggested models 

saimedhi:
### Description

- Added popover to display links to suggested models 
- Included both Embedding and Large Language model examples. 

[screen-capture (41).webm](https://github.com/user-attachments/assets/6ea9cad6-adc2-499e-b163-ed6730e8d538)

Screenshot 2025-02-12 at 10 16 52 AM

Screenshot 2025-02-12 at 10 16 12 AM

### Check List

### DIFF
--- a/common/constants.ts
+++ b/common/constants.ts
@@ -259,6 +259,8 @@ export const SEARCH_PIPELINE_DOCS_LINK =
   'https://opensearch.org/docs/latest/search-plugins/search-pipelines/using-search-pipeline/';
 export const ML_RESPONSE_PROCESSOR_EXAMPLE_DOCS_LINK =
   'https://opensearch.org/docs/latest/search-plugins/search-pipelines/ml-inference-search-response/#example-externally-hosted-text-embedding-model';
+export const UPDATE_MODEL_DOCS_LINK =
+  'https://opensearch.org/docs/latest/ml-commons-plugin/api/model-apis/update-model/';
 
 /**
  * Text chunking algorithm constants
--- a/public/pages/workflow_detail/workflow_inputs/input_fields/model_field.tsx
+++ b/public/pages/workflow_detail/workflow_inputs/input_fields/model_field.tsx
@@ -18,6 +18,7 @@ import {
   EuiSmallButtonIcon,
   EuiFlexGroup,
   EuiFlexItem,
+  EuiHealth,
 } from '@elastic/eui';
 import {
   MODEL_STATE,
@@ -25,6 +26,7 @@ import {
   ModelFormValue,
   ML_CHOOSE_MODEL_LINK,
   FETCH_ALL_QUERY_LARGE,
+  UPDATE_MODEL_DOCS_LINK,
 } from '../../../../../common';
 import { AppState, searchModels, useAppDispatch } from '../../../../store';
 import { getDataSourceId } from '../../../../utils';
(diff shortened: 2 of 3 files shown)


common/constants.ts
```diff
@@ -253,6 +253,23 @@ export const SEARCH_PIPELINE_DOCS_LINK =
...
+export const DEEPSEEK_CHAT_DOCS_LINK = 
+  'https://github.com/opensearch-project/ml-commons/blob/main/docs/remote_inference_blueprints/deepseek_connector_chat_blueprint.md';
+
+// Embedding Models Documentation Links
+export const COHERE_EMBEDDING_MODEL_DOCS_LINK = 
+  'https://github.com/opensearch-project/ml-commons/blob/2.x/docs/remote_inference_blueprints/cohere_connector_embedding_blueprint.md';
+
+export const BEDROCK_TITAN_EMBEDDING_DOCS_LINK = 
+  'https://github.com/opensearch-project/ml-commons/blob/2.x/docs/remote_inference_blueprints/bedrock_connector_titan_embedding_blueprint.md';
+
```
ohltyler, line 272: These are the blueprints we want to avoid customers visiting, as they are not compatible with ML inference processors, and generally do not have model interfaces, which means they are not compatible with OpenSearch Flow. We need to point to https://github.com/opensearch-project/dashboards-flow-framework/blob/main/documentation/models.md

For any models not present in that file ^, then suggest to add.

public/pages/workflow_detail/workflow_inputs/input_fields/model_field.tsx
```diff
@@ -125,17 +133,58 @@ export function ModelField(props: ModelFieldProps) {
             (props.showInvalid ?? true) &&
             getIn(errors, `${field.name}.id`) &&
             getIn(touched, `${field.name}.id`);
+          const renderPopover = () => {
+            if (props.fieldPath === "llm") {
```
ohltyler, line 137: This isn't a reliable way to determine whether or not we should suggest a model of a certain type; field path names are internal and may change. Suggest to add a prop to `ModelField` that passes an optional model type, for example. Maybe make an enum of different model types, which is currently just LLM/generative models, and embedding models.

public/pages/workflow_detail/workflow_inputs/input_fields/model_field.tsx
```diff
@@ -61,6 +63,12 @@ export function ModelField(props: ModelFieldProps) {
   // Deployed models state
   const [deployedModels, setDeployedModels] = useState<ModelItem[]>([]);
 
+  // Embedding model popover state
+  const [embeddingModelPopoverOpen, setEmbeddingModelPopoverOpen] = useState<boolean>(false);
+
+  // Large Language model popover state
+  const [largeLaguageModelPopoverOpen, setLargeLaguageModelPopoverOpen] = useState<boolean>(false);
```
ohltyler, line 70: How about just have a single generic popover state?

public/pages/workflows/new_workflow/embedding_model_popover_content.tsx
ohltyler, line 1: These components are very repetitive. How about we have a single component, that takes in a single prop determining the model type, and then dynamically render? They seem identical besides the list of models to display, and maybe one-word differences, like "embedding" vs. "large language".

Additionally, this component should just live next to `model_field` where it is logically coupled, instead of nested under `new_workflow`.

ohltyler (changes requested):

PR #646, opened 2025-02-26
Fixed local cluster bug and added datasource version compatibility check before importing workflows

saimedhi:
### Description

- Fixed local cluster bug and added datasource version compatibility check before importing workflows
- I will add testing video

### Issues Resolved

[List any issues this PR will resolve]

### Check List

### DIFF
--- a/public/pages/workflows/import_workflow/import_workflow_modal.tsx
+++ b/public/pages/workflows/import_workflow/import_workflow_modal.tsx
@@ -27,6 +27,7 @@ import {
   getObjFromJsonOrYamlString,
   isValidUiWorkflow,
   isValidWorkflow,
+  isCompatibleWorkflow,
 } from '../../../utils';
 import { getCore } from '../../../services';
 import {
@@ -86,6 +87,11 @@ export function ImportWorkflowModal(props: ImportWorkflowModalProps) {
     return description.length > MAX_DESCRIPTION_LENGTH;
   }
 
+  // States for tracking workflow template compatibility with current data source version
+  const [isCompatible, setIsCompatible] = useState<boolean>(true);
+  const [inCompatibleDataSourceVersion, setInCompatibleDataSourceVersion] = useState<string | undefined>();
+
+
   // transient importing state for button state
   const [isImporting, setIsImporting] = useState<boolean>(false);
 
@@ -115,6 +121,19 @@ export function ImportWorkflowModal(props: ImportWorkflowModalProps) {
     }
   }, [fileObj]);
 
+  useEffect(() => {
+    async function checkCompatibility() {
+      if (isValidWorkflow(fileObj)) {
+        const [isCompatible, dataSourceVersion] = await isCompatibleWorkflow(fileObj, dataSourceId);
+        setIsCompatible(isCompatible);
+        if (dataSourceVersion) {
+          setInCompatibleDataSourceVersion(dataSourceVersion);
+        }
+      }
+    }
+    checkCompatibility();
+  }, [fileObj, dataSourceId]);
+
   function onModalClose(): void {
     props.setIsImportModalOpen(false);
     setFileContents(undefined);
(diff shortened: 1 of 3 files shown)


public/utils/utils.tsx
```diff
@@ -157,6 +157,30 @@ export function isValidWorkflow(workflowObj: any): boolean {
...
+  const [effectiveMajorVersion, effectiveMinorVersion] = dataSourceVersion.split('.').map(Number);
+  
+  // Checks if any version in compatibility array matches the current dataSourceVersion (major.minor)
+  const isCompatible = compatibility.some(compatibleVersion => {
+    const [compatibleMajor, compatibleMinor] = compatibleVersion.split('.').map(Number);
+    return effectiveMajorVersion === compatibleMajor && effectiveMinorVersion === compatibleMinor;
+  });
+
+  return [isCompatible, `${effectiveMajorVersion}.${effectiveMinorVersion}`];
+}
```
ohltyler, line 181: This function seems overly complex. I think it can be simplified to just checking compatibility and returning a boolean. In the case of it being incompatible, it means there is a current data source, and we can just use the existing util `getEffectiveVersion` to get the version. Then you don't need to pass around multiple vars, return `undefined` on line 168, etc.

public/pages/workflows/import_workflow/import_workflow_modal.tsx
```diff
@@ -86,6 +87,11 @@ export function ImportWorkflowModal(props: ImportWorkflowModalProps) {
     return description.length > MAX_DESCRIPTION_LENGTH;
   }
 
+  // States for tracking workflow template compatibility with current data source version
+  const [isCompatible, setIsCompatible] = useState<boolean>(true);
+  const [inCompatibleDataSourceVersion, setInCompatibleDataSourceVersion] = useState<string | undefined>();
```
ohltyler, line 92: similar to other comment, this state var shouldn't be necessary. Can just fetch the data source version after getting the data source ID (under line 64), and if incompatible, include in the callout directly.

ohltyler (changes requested):

public/utils/utils.tsx
saimedhi: - I initially thought about doing the same. but since in the error message i am only displaying `${effectiveMajorVersion}.${effectiveMinorVersion}` implies when datasource version is `3.0.0 alpha 1` , we display `3.0`. 

    These small steps are already done in above function. So, I thought it will prevent duplicate code.  

- I can do as per your suggestion. But please confirm if you are good with below error message having `3.0` instead of `3.0.0 alpha 1`

`The uploaded file is not compatible with the current data source version 3.0. Upload a compatible file or switch to another data source.`

ohltyler: Version suffixes will be interim and not included for any actual releases, so that is fine. But, if you still want to handle this case, then how about enhancing `getEffectiveVersion()` fn or just add some extra post-processing when formulating the modal text? Either way, the version logic can/should be removed from this util fn.

saimedhi: **Additional Context:** 
In below example -  import workflow template,  Datasources with versions 2.19.x and 3.0.x are compatible for importing this workflow. 

```
 {
  "name": "import workflow",
  "use_case": "",
  "description": "",
  "version": {
    "template": "1.0.0",
    "compatibility": [
      "2.19.0",
      "3.0.0"
    ]
  },
  "workflows": {....
```

saimedhi: Made changes as suggested. Created an other utils function to format versions. Instead of disturbing getEffectiveVersion.

saimedhi: Tested  again after changes: 
Screenshot 2025-02-26 at 1 25 33 PM

ohltyler (approved):
LGTM! Great improvements.

PR #664, opened 2025-03-06
Update plugin name and rearrange Try AI Search Flows card

saimedhi:
### Description

- Update plugin card in Search Overview Page and rearrange "Try AI Search Flows card". 
- **Please recheck when backporting. `Removed "plugin_version": "2.19.0.0" in package.json`. Fixed yarn build script accordingly.** If needed will raise separate PR for main. 

- Relavant PR in OSD https://github.com/opensearch-project/OpenSearch-Dashboards/pull/9500

### Check List

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "OpenSearch Flow Framework Dashboards Plugin",
   "main": "index.js",
   "config": {
-    "plugin_version": "2.17.0.0",
+    "plugin_version": "2.19.0.0",
     "plugin_name": "flowFrameworkDashboards",
     "plugin_zip_name": "dashboards-flow-framework"
   },
--- a/package.json
+++ b/package.json
@@ -4,7 +4,6 @@
   "description": "OpenSearch Flow Framework Dashboards Plugin",
   "main": "index.js",
   "config": {
-    "plugin_version": "2.19.0.0",
     "plugin_name": "flowFrameworkDashboards",
     "plugin_zip_name": "dashboards-flow-framework"
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "OpenSearch Flow Framework Dashboards Plugin",
   "main": "index.js",
   "config": {
-    "plugin_version": "2.17.0.0",
```
ohltyler, line 7: do we need this config field at all? [AD](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/blob/main/package.json) does not have it anymore for example. Not sure if it's used in the build script- my assumption is not, since this is out-of-date and hasn't had impact thus far.

public/general_components/service_card/plugin_card.tsx
```diff
@@ -37,7 +37,7 @@ export const registerPluginCard = (
           }}
         >
           {i18n.translate('flowFrameworkDashboards.opensearchFlowCard.footer', {
-            defaultMessage: 'Try OpenSearch Flow',
+            defaultMessage: 'Try AI Search Flows',
```
ohltyler, line 40: Can you replace with the constant `PLUGIN_NAME`? Then this change won't be necessary in the future if it updates again.

package.json
saimedhi: While building zip file this version is used for naming build zip. Will check by removing.

saimedhi: Fixed Build script. Removed `"plugin_version": "2.17.0.0"`

ohltyler: Awesome thanks for the fix!

ohltyler (approved):
LGTM!

PR #676, opened 2025-03-13
Simplify ML processor form when interface is defined

ohltyler:
### Description

Simplifies the controls & form components for ML processors when the underlying model has a defined interface. Make the inputs & outputs readonly, and prevent adding/deleting entries. The main idea, is that if a model has a defined interface, add more guardrails to prevent user confusion from creating invalid/unnecessary mappings.

### Issues Resolved

N/A

### Check List


saimedhi (approved):
LGTM

PR #685, opened 2025-03-27
Bump version to 3.0.0.0-beta1

ohltyler:
### Description

Bump version to 3.0.0.0-beta1

### Issues Resolved

Makes progress on #304 

### Check List

### DIFF
--- a/opensearch_dashboards.json
+++ b/opensearch_dashboards.json
@@ -1,7 +1,7 @@
 {
   "id": "flowFrameworkDashboards",
-  "version": "3.0.0.0-alpha1",
-  "opensearchDashboardsVersion": "3.0.0-alpha1",
+  "version": "3.0.0.0-beta1",
+  "opensearchDashboardsVersion": "3.0.0",
   "server": true,
   "ui": true,
   "requiredBundles": [],
--- a/.github/workflows/remote-integ-tests-workflow.yml
+++ b/.github/workflows/remote-integ-tests-workflow.yml
@@ -6,8 +6,8 @@ env:
   CI: 1
   TERM: xterm
   OPENSEARCH_DASHBOARDS_VERSION: 'main'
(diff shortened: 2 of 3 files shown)


opensearch_dashboards.json
```diff
@@ -1,7 +1,7 @@
 {
   "id": "flowFrameworkDashboards",
-  "version": "3.0.0.0-alpha1",
-  "opensearchDashboardsVersion": "3.0.0-alpha1",
+  "version": "3.0.0.0-beta1",
+  "opensearchDashboardsVersion": "3.0.0",
```
saimedhi, line 4: @ohltyler,OSD version should be "3.0.0-beta1"

saimedhi (changes requested):

opensearch_dashboards.json
ohltyler: The early recommendations actually had mentioned to not make this change, and either way the tests are passing. Suggest we leave as-is, additionally it means less changes for future 3.0 incremental releases.

ohltyler: Any other reason you think it is needed?

saimedhi: Nope. I see that in OSD main branch version is "3.0.0"

https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/package.json#L14

saimedhi: @ohltyler , Also one more thing. Unrelated to above. Lets start adding changelog for each PR. So, that for release notes we can just copy from changelog.

saimedhi (approved):

opensearch_dashboards.json
ohltyler: Agreed- forgot that it was added, but yes this can make things easier.

PR #692, opened 2025-04-02
disable choosing index in search pipeline if index is ingested in ingest pipeline

kamahuan:
### Description
If the user has already ingested an index in the ingest pipeline, in the search pipeline, should disable the index drop down list so that user can use the index in the ingest pipeline.

### Issues Resolved

NA

### Check List

### DIFF
--- a/public/pages/workflow_detail/workflow_inputs/processors_list.tsx
+++ b/public/pages/workflow_detail/workflow_inputs/processors_list.tsx
@@ -94,6 +94,11 @@ export function ProcessorsList(props: ProcessorsListProps) {
     [processorId: string]: boolean;
   }>({});
 
+  // check if an index has already been indexed in the ingest pipeline
+  const hasIngestedIndex =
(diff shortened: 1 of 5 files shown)


public/pages/workflow_detail/workflow_inputs/processors_list.tsx
```diff
@@ -94,6 +94,11 @@ export function ProcessorsList(props: ProcessorsListProps) {
     [processorId: string]: boolean;
   }>({});
 
+  // check if an index has already been indexed in the ingest pipeline
+  const hasIngestedIndex =
+    props.uiConfig?.ingest?.index?.name?.value !== undefined &&
+    props.uiConfig?.ingest?.index?.name?.value !== '';
```
jackiehanyang, line 100: Is there another way to check this instead of relying on the prop's value? It assumes that the presence of this value means the index has been ingested, which might not always be true. For example, if the upstream component has issues, it could affect the prop value being passed down.

kamahuan:
Will add a new PR related to this issue.

PR #694, opened 2025-04-08
disable index in search pipeline if already indexed in ingest pipeline

kamahuan:
### Description

If the user ingest an index in the ingest pipeline, the index should be reflected in the search pipeline

### Issues Resolved

NA

### Demo

https://github.com/user-attachments/assets/c00462b1-c538-4556-91f1-d86bb418eeeb

### Check List


ohltyler:
The user journey of enhancing an existing index will require using the reindex API and further UX work to finalize that, which is in progress. We do not want to start ingesting data on top of an existing index, which is what this change is doing. The ingest flow still & will always require ingesting data to a _new_ index. As a reminder, the "from existing index" in the data source selector is just another option for gathering some sample data to _add_ to this new index.

PR #707, opened 2025-04-22
Consolidate ingest response and errors into the console section

kamahuan:
### Description

Consolidate ingest response and errors into the console section and place the section at the bottom of the workflow page. Upon discussion with UX team, the console section will push other panels up when expanded instead of overlaying other panels.

### Issues Resolved
NONE

### Demo video

https://github.com/user-attachments/assets/9d1b6f9e-67f4-4c98-9e2d-72321b50d7e2

### 2nd Revision Changes:
1. Increase the height of the console section when not expanded.
2. Remove the horizontal bar in the console section. This is not needed since console is expanding  full width horizontally. The other vertical scroll bars are kept since pushing the console section above will hide upper panel content.
3. Refactor the console section into a component with 2 child components.
4. Use enums already defined in the constant file instead of creating new variables

### Check List

### DIFF
--- a/public/pages/workflow_detail/tools/tools.tsx
+++ b/public/pages/workflow_detail/tools/tools.tsx
@@ -20,8 +20,8 @@ import {
   CONFIG_STEP,
   customStringify,
   FETCH_ALL_QUERY,
+  getVisibleTabs,
   INSPECTOR_TAB_ID,
-  INSPECTOR_TABS,
   QueryParam,
   SearchResponse,
   Workflow,
@@ -43,8 +43,6 @@ interface ToolsProps {
   selectedTabId: INSPECTOR_TAB_ID;
   setSelectedTabId: (tabId: INSPECTOR_TAB_ID) => void;
   selectedStep: CONFIG_STEP;
-  hideIngestResponseTab?: boolean;
-  hideErrorsTab?: boolean;
 }
 
 const PANEL_TITLE = 'Inspect flows';
@@ -66,12 +64,7 @@ export function Tools(props: ToolsProps) {
   >([]);
   const { values } = useFormikContext<WorkflowFormValues>();
 
-  const visibleTabs = INSPECTOR_TABS.filter((tab) => {
-    if (props.hideIngestResponseTab && tab.id === INSPECTOR_TAB_ID.INGEST)
-      return false;
-    if (props.hideErrorsTab && tab.id === INSPECTOR_TAB_ID.ERRORS) return false;
-    return true;
-  });
+  const visibleTabs = getVisibleTabs();
 
   // Standalone / sandboxed search request state. Users can test things out
   // without updating the base form / persisted value.
@@ -132,14 +125,14 @@ export function Tools(props: ToolsProps) {
 
   // auto-navigate to errors tab if new errors have been found
   useEffect(() => {
-    if (curErrorMessages.length > 0 && !props.hideErrorsTab) {
+    if (curErrorMessages.length > 0) {
       props.setSelectedTabId(INSPECTOR_TAB_ID.ERRORS);
     }
   }, [curErrorMessages]);
 
   // auto-navigate to ingest tab if a populated value has been set, indicating ingest has been ran
   useEffect(() => {
-    if (!isEmpty(props.ingestResponse) && !props.hideIngestResponseTab) {
+    if (!isEmpty(props.ingestResponse)) {
       props.setSelectedTabId(INSPECTOR_TAB_ID.INGEST);
     }
   }, [props.ingestResponse]);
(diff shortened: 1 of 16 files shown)


public/pages/workflow_detail/tools/tools.tsx
```diff
@@ -43,6 +43,8 @@ interface ToolsProps {
   selectedTabId: INSPECTOR_TAB_ID;
   setSelectedTabId: (tabId: INSPECTOR_TAB_ID) => void;
   selectedStep: CONFIG_STEP;
+  hideIngestResponseTab?: boolean;
+  hideErrorsTab?: boolean;
```
ohltyler, line 47: Why do we need these flags? Is there a situation where we conditionally show or don't show the tabs? If not, suggest we just update what the tabs are in the respective components directly.

public/pages/workflow_detail/workflow_inputs/ingest_inputs/advanced_settings.tsx
```diff
@@ -28,12 +28,15 @@ import {
 
 interface AdvancedSettingsProps {
   setHasInvalidDimensions: (hasInvalidDimensions: boolean) => void;
+  onToggle?: (isExpanded: boolean) => void;
```
ohltyler, line 31: what is the purpose of this prop? Is the idea to collapse all accordions if the bottom console is opened?

public/pages/workflow_detail/workflow_inputs/workflow_inputs.tsx
```diff
@@ -84,6 +89,11 @@ interface WorkflowInputsProps {
   setUnsavedSearchProcessors: (unsavedSearchProcessors: boolean) => void;
   displaySearchPanel: () => void;
   setCachedFormikState: (cachedFormikState: CachedFormikState) => void;
+  context: 'INGEST' | 'SEARCH';
```
ohltyler, line 92: Please add as a constant in `common/constants`

public/pages/workflow_detail/workflow_inputs/workflow_inputs.tsx
```diff
@@ -269,6 +285,119 @@ export function WorkflowInputs(props: WorkflowInputsProps) {
...
+  useEffect(() => {
+    if (props.isConsoleExpanded !== undefined) {
+      setConsoleExpanded(props.isConsoleExpanded);
+    }
+  }, [props.isConsoleExpanded]);
+
+  useEffect(() => {
+    if (props.setConsoleContent) {
+      const consoleContent = (
+        <EuiFlexGroup direction="column" gutterSize="s">
```
ohltyler, line 304: Strongly suggest to move all console content into a single parent component, and child sub-components, similar to the existing `<Tools />` component, where each tab is a child component. You should be able to reuse most/all of the logic in the existing [Errors](https://github.com/opensearch-project/dashboards-flow-framework/blob/main/public/pages/workflow_detail/tools/errors/errors.tsx) and [ingest](https://github.com/opensearch-project/dashboards-flow-framework/blob/main/public/pages/workflow_detail/tools/ingest/ingest.tsx) components for rendering the content within the console.

public/pages/workflow_detail/workflow_inputs/workflow_inputs.tsx
ohltyler: in fact, there is an existing `CONFIG_STEP` enum which you can probably use.

public/pages/workflow_detail/tools/tools.tsx
kamahuan: According to the new mockup, we will not need ingest response and errors tab, so they will always be false, will remove this part

public/pages/workflow_detail/workflow_inputs/ingest_inputs/advanced_settings.tsx
kamahuan: When I initially testing the styles, it looks like if the console is expanded, it will hide `advanced settings` so I add this prop to track if the accordion is expanded or collapsed, and correspondingly make advanced settings to be visible. Tyler do you think if it is appropriate to adjust the styling like this? Do you have any other recommendations?

ohltyler: > it looks like if the console is expanded, it will hide advanced settings so I add this prop to track if the accordion is expanded or collapsed, and correspondingly make advanced settings to be visible

Not sure I follow. If the console is pushing up the other components and not overlaying, nothing should be getting hidden. In general, suggest we start simple, and not have other side effects / unnecessary listeners on deeply nested components like the ingest's advanced settings accordion, until we feel it's needed.

public/pages/workflow_detail/workflow_inputs/workflow_inputs.tsx
kamahuan: Gotcha, will use `CONFIG_STEP`

ohltyler (commented):
Done another pass on the latest changes. In general, I think it's worth understanding the existing `Tools` component logic more, and maybe start by refactoring that to create the new `Console` component. The current `Console` component has some different behavior and unnecessary complication.

kamahuan:
Rewrite the console component based on the ingest and error component to make it cleaner. Please check: https://github.com/opensearch-project/dashboards-flow-framework/pull/730

PR #724, opened 2025-05-14
Refactor form navigation to left panel

ohltyler:
### Description

Refactors the workflow detail page, such that users click on a component in the new left nav panel, and configure in the central panel. All components (ingest and search) are viewable altogether from the new left nav. The same guardrails are in place, such that users cannot make changes to the search flow if there are unsaved changes to ingest, and vice versa.

Main component changes:
- `LeftNav`: the base component with all of the components
- `IngestContent`: the accordion of all of the ingest-flow-related components
- `SearchContent`: the accordion of all of the search-flow-related components
- `ComponentInput`: the rough "replacement" of `WorkflowInputs`. Dynamically renders based on the selected component from the left nav. Shows the relevant form input for such selected component.

Demo video showing the new pattern, new form states, and overall user flow:

[left-nav-demo.webm](https://github.com/user-attachments/assets/77b83e9e-cabd-4a7b-9544-f67a96e2ecae)

### Check List


kamahuan:
Hey Tyler, I downloaded the code and noticed a few issues. Please correct me if my understanding is not right:
image

1. If the ingest flow has any errors, the mockup says ingest flow will roll back to `Not created` state, but the UI shows `active` state.
2.  Wording on the right panel under `Inspect` is inconsistent with the mockup.
3. `test flow` tab on the right panel always has a `run test` button and the UI shows more context than the mockup.
4. The icon on the upper top left panel seems different from the mockup.

Another problem is that should we add a button in the sample data section so that user can delete the sample data if they change their mind. Currently there is no button to roll back to the initial state.

image
The second screenshots show I created ingest sample data and create the ingest pipeline, and then change to another sample data and config the processor, do we need `create ingest pipeline` button in this case? I can see the `update` button shows if I make any changes to the processor. The red box is inconsistent wording and icon.

image
The third image shows that if the user clicked any component, it will show `selected`. Did not see the same `selected` text on the mockup. 
Tried click `Retrive from data source` after search pipeline created but returned nothing. Is this behavior expected.

image
Another issue is that when there is unsaved changes, the mockup shows a callout instead of the `update` button.

ohltyler:
Thanks for taking a look. As a reminder, this is to get the bulk of the functionality into the feature branch. The final styling/wording/spacing etc is not finalized yet.

> If the ingest flow has any errors, the mockup says ingest flow will roll back to Not created state, but the UI shows active state.

It greatly depends on what the error is. If it is an error creating, it will stay in not created state. If it is an error running ingest, it will still be active, as it just means there was some processor causing some failure and should be updated, but the resources are still active/created. This is the expected behavior.

> Wording on the right panel under Inspect is inconsistent with the mockup.

We can ignore final details/wording at this stage, this will be done in later UX review finalization stages.

> test flow tab on the right panel always has a run test button and the UI shows more context than the mockup.

This is the expected behavior, mockups may not show everything.

> The icon on the upper top left panel seems different from the mockup.

Same comment as previous, this will all be finalized later on. I chose this one over the mockup due to sizing issues currently.

> The second screenshots show I created ingest sample data and create the ingest pipeline, and then change to another sample data and config the processor, do we need `create ingest pipeline` button in this case? I can see the `update` button shows if I make any changes to the processor.

Yes, because if any changes to the ingest pipeline, the data needs to be re-ingested to be processed correctly.

> The red box is inconsistent wording and icon.

This is a moving target from UX end and still being finalized, current implementation matches previous mocks. This like above callouts around small wording/styling will be finalized later on.

> The third image shows that if the user clicked any component, it will show `selected`. Did not see the same `selected` text on the mockup. Tried click `Retrive from data source` after search pipeline created but returned nothing. Is this behavior expected.

Same comments as above, this is still being finalized. 

> Another issue is that when there is unsaved changes, the mockup shows a callout instead of the `update` button.

Same comments as above, this is still being finalized.

saimedhi:
Hi Tyler, thank you for the PR! I'll start reviewing it now.

saimedhi (approved):
@ohltyler, overall looks great to me. This PR looks good to be merged into feature branch. 

I will do more testing in all different scenarios  and let you know if I find any bug. Overall nav_components and component_input code looks clean

PR #757, opened 2025-07-28
Cleanly handle error states for backend connection issues

ohltyler:
### Description

Improves the UI to cleanly handle scenarios where there are failures to the cluster (most commonly the backend plugin APIs), whether it is a local cluster or external data source (if MDS enabled). Before, there would be ad-hoc toast messages indicating the failures, but no limitation to navigating around the plugin, which could lead to weird side effects (buttons not doing anything, creation failing, etc.) when there was issues executing the plugin APIs. Now, we have blanket messaging on the page to prevent users from triggering these states altogether, and more obvious direction for the users to first fix their cluster before coming back to the UI.

The 2 caught scenarios are:
**Scenario 1.** Connection issues + MDS disabled: provide wording indicating failures accessing the cluster, and action wording to make sure the cluster has all necessary plugins installed.

**Scenario 2.** Connection issues + MDS enabled: provide wording indicating failures accessing the cluster, and action wording to make sure the cluster is available and has all necessary features available. _Note: there may not be sufficient metadata in the configured data sources to know exactly why some API calls may be failing, so we keep the wording generic. We guide the users by directing them to the data source configuration for further debugging._

Screenshots:
TODO

### Check List


public/pages/workflows/new_workflow/quick_configure_modal.tsx
```diff
@@ -330,56 +334,57 @@ export function QuickConfigureModal(props: QuickConfigureModalProps) {
                             <EuiSpacer size="s" />
                           </>
                         )}
-                        {props.workflow?.ui_metadata?.type === WORKFLOW_TYPE.SEMANTIC_SEARCH_USING_SPARSE_ENCODERS ? (
```
ohltyler, line 333: There is no changes on lines 333-387; auto-formatting updated this part.

ohltyler:
Verified integ tests pass locally. Failing on CI due to 3.1/3.2 version mismatch, will re-run after merging that PR.
```
  (Run Finished)


       Spec                                              Tests  Passing  Failing  Pending  Skipped  
  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ ✔  plugins/dashboards-flow-framework/c      02:43        6        6        -        -        - │
  │    reate_workflow_spec.js                                                                      │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘
    ✔  All specs passed!                        02:43        6        6        -        -        -  
```

public/pages/workflows/workflows.tsx
```diff
@@ -99,6 +106,45 @@ export function Workflows(props: WorkflowsProps) {
...
+    async function mlCommonsHealthCheck() {
+      await dispatch(
+        searchModels({
+          apiBody: FETCH_ALL_QUERY_LARGE,
+          dataSourceId,
+        })
+      ).then((resp: any) => {
+        setMLCommonsConnectionErrors(!isEmpty(resp.error));
+      });
+    }
```
jackiehanyang, line 141: nit: it seems like you're making two independent async calls here, maybe consider run them in parallel for efficiency

public/pages/workflows/workflows.tsx
```diff
@@ -277,7 +323,55 @@ export function Workflows(props: WorkflowsProps) {
...
+                title={<h2>Error accessing cluster</h2>}
+                body={
+                  <p>
+                    Ensure your OpenSearch cluster is available and has the Flow
+                    Framework and ML Commons plugins installed.
+                  </p>
+                }
+                actions={
+                  <EuiButton color="primary" fill={false}>
+                    <EuiLink target="_blank" href={MAIN_PLUGIN_DOC_LINK}>
```
jackiehanyang, line 342: EuiButton already supports `href` directly, no need to nest it with EuiLink? Line 364 below has the correct usage

public/pages/workflows/workflows.tsx
ohltyler: Thanks for the suggestion - updated to run them inside a single `Promise.all()` list.

public/pages/workflows/workflows.tsx
ohltyler: Originally changed since I needed the popout icon since I've set `target=_blank`. Utilized icon params exposed from `EuiButton` instead and removed `EuiLink`, thanks for the suggestion!

jackiehanyang (approved):

PR #787, opened 2025-10-01
Add more guardrails based on agent types

ohltyler:
### Description

Misc guardrail enhancements:
- removes the 'optional' field for memory, and sets the default memory configuration when creating a new conversational agent, as such it's impossible to create a new conversational agent _without_ memory configured now.
- removes the "All indices" option in the index selection for flow agents. If a flow agent is selected, users must select an index. If a conversational index is selected, it is defaulted to "all indices". If a user switches from flow -> conversational, the selected index remains the same. If a user switches from conversational -> flow, and if "all indices" was selected, it is set back to blank/empty.

### Check List

### DIFF
--- a/public/pages/workflow_detail/agentic_search/test_flow/index_selector.tsx
+++ b/public/pages/workflow_detail/agentic_search/test_flow/index_selector.tsx
@@ -25,13 +25,16 @@ import {
 } from '../../../../store';
 import { getDataSourceId } from '../../../../utils';
 import {
+  AGENT_TYPE,
   OMIT_SYSTEM_INDEX_PATTERN,
   WorkflowFormValues,
 } from '../../../../../common';
 import { IndexDetailsModal } from './index_details_modal';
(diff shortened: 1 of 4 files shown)


public/pages/workflow_detail/agentic_search/test_flow/index_selector.tsx
```diff
@@ -55,20 +58,28 @@ export function IndexSelector(props: IndexSelectorProps) {
...
+      eligibleIndexOptions = [
+        {
+          text: ALL_INDICES,
+          value: '',
+        },
+        ...eligibleIndexOptions,
+      ];
+    }
+    setIndexOptions(eligibleIndexOptions);
+  }, [indices, props.agentType]);
```
owaiskazi19, line 82: Do we want to have a test for this change?

ohltyler: Good idea, added in latest commit.

PR #793, opened 2025-10-02
Improve error handling from server-side

### LINKED ISSUE #788: [BUG] Docker distributions don't set API failures as `rejected` in redux
Certain redux store variables, relied upon to be safe, are getting corrupted and set to empty values like `undefined`. This can cause other side effects and NPEs, potentially putting the UI in odd error states.

Currently, only able to reproduce in docker distributions. The root cause is related to the API calls not being set as `rejected` if they fail for whatever reason (e.g., searching for agents when the system index hasn't been initialized). In a local environment, this is caught, set as `rejected`, and the error is propagated. In a docker environment, it is being set as `fulfilled` state in redux, even if empty, leading to the corruption and downstream NPEs.

Further research is needed to identify the nuance.

ohltyler:
### Description

Changes the error handling pattern for server-side errors. Rather than propagating strings and checking for errors based on `instanceof HttpFetchError`, we throw the errors directly, and catch them to be rejected in redux. This prevents issues under certain bundled builds, where the `instanceof` check will fail, and an error will be swallowed, leaving to corrupt redux store variables and unexpected NPEs.

### Issues Resolved

Closes #788 

### Check List

### DIFF
--- a/public/route_service.ts
+++ b/public/route_service.ts
@@ -4,7 +4,7 @@
  */
 
 import { isEmpty } from 'lodash';
-import { CoreStart, HttpFetchError } from '../../../src/core/public';
+import { CoreStart } from '../../../src/core/public';
 import {
   CREATE_WORKFLOW_NODE_API_PATH,
   DELETE_WORKFLOW_NODE_API_PATH,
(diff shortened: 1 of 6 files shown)


public/route_service.ts
```diff
@@ -332,7 +289,7 @@ export function configureRoutes(core: CoreStart): RouteService {
         );
         return response;
       } catch (e: any) {
```
amitgalitz, line 291: why not return something more defined, for example:
```
return opensearchDashboardsResponse.ok({
        body: {
          ok: false,
          error: getErrorMessage(err),
        },
      });
      ```

ohltyler: The issue lies in the typing of the errors returned from server-side, in bundled vs. local environments. The content is still consistent, as everything is formatted the same from server-side. So, by throwing this error, it avoids any rigidity and ensures that any/all errors vs. non-errors are handled

ohltyler: Those response types are all defined server-side.

amitgalitz: I get that but I feel like if we are catching something here we can just do the formatting here and return the response like this
its not a big issue, your way is okay too, i am just used to something like that from AD mostly

amitgalitz (approved):

public/route_service.ts
ohltyler: We do have all of that formatting on the server-side, here's the [flow framework APIs](https://github.com/opensearch-project/dashboards-flow-framework/blob/main/server/routes/flow_framework_routes_service.ts) for example. The change here is to remove the explicit type checking and pass as `any`, such that the `instanceof` checks and bugs will be resolved.

The last bit of formatting, like adding the prefix, was the existing pattern, and I kept the same.

PR #872, opened 2026-04-03
Add fallback query configuration for QueryPlanningTool

### LINKED ISSUE #868: [Agentic Search] Add fallback query configuration for QueryPlanningTool
## Description

The QueryPlanningTool now supports configuring a fallback query, ensuring that if the primary agentic query fails or returns no results, a backup query is automatically triggered. The frontend should expose this configuration option.

## Current Behavior

The UI does not provide a way to configure a fallback query for the QueryPlanningTool. If the LLM-generated query fails or returns empty results, users see no results with no recovery mechanism.

## Expected Behavior

- When configuring the QueryPlanningTool in the agent configuration panel, users should be able to optionally define a fallback query (as a JSON DSL query body)
- The fallback query is used when the LLM-generated query fails to parse or returns no results
- The UI should provide a JSON editor for the fallback query with validation
- The default fallback (`{"size":10,"query":{"match_all":{}}}`) should be shown as a placeholder or default value

## Backend Reference

The `QueryPlanningTool` includes a default fallback query `{"size":10,"query":{"match_all":{}}}`. When no valid JSON can be extracted from the LLM response, the tool returns this fallback. Users can configure a custom fallback query.

## References

- [QueryPlanningTool fallback behavior](https://docs.opensearch.org/latest/ml-commons-plugin/agents-tools/tools/query-planning-tool/#fallback-behavior)
- Meta issue: #865

ohltyler:
### Description

Adds an optional fallback query configuration field to the QueryPlanningTool in the agentic search UI. When the LLM-generated query fails to parse or returns no results, the backend uses this fallback query instead.

### Changes

**UI** (`public/pages/.../tool_forms/query_planning_tool.tsx`):
- Added `SimplifiedJsonField` (JSON code editor with syntax highlighting and auto-formatting) for the `fallback_query` tool parameter
- Field appears below the generation type config in the QueryPlanningTool accordion
- Value is persisted as a string directly to avoid `parseStringOrJson` converting it to an object (which crashes the Ace editor)

### Issues Resolved

Resolves #868

### Testing

**Unit tests:** All 126 tests pass (23 suites), no regressions.

**End-to-end validation:**
1. Created a flow agent with `QueryPlanningTool` backed by a Bedrock Claude model
2. Configured a custom fallback query via the JSON editor in the UI (e.g., a fuzzy multi_match on title/description using `${parameters.question}`)
3. Verified the fallback query persists in the agent config (`GET /_plugins/_ml/agents/<id>`)
4. Triggered the fallback by asking a question completely unrelated to the index mapping (e.g., "Find all employees hired in 2023" against an ecommerce index) — the LLM generates a query with non-existent fields, causing the fallback to activate
5. Confirmed the fallback query executes and returns relevant results instead of empty results

Demo:

[screen-capture.webm](https://github.com/user-attachments/assets/b54b1cdf-810c-4374-a022-9f8896c0c364)

Error handling (invalid JSON) demo:

[screen-capture (1).webm](https://github.com/user-attachments/assets/a2880a02-278d-4c19-8d64-331bb4b137bd)

### Check List
- [x] New functionality includes testing
- [x] New functionality has been documented

### DIFF
--- a/public/pages/workflow_detail/agentic_search/configure_flow/agent_tools.test.tsx
+++ b/public/pages/workflow_detail/agentic_search/configure_flow/agent_tools.test.tsx
@@ -4,7 +4,7 @@
  */
 
 import React from 'react';
-import { render, screen } from '@testing-library/react';
+import { render, screen, act } from '@testing-library/react';
 import '@testing-library/jest-dom';
 import { Provider } from 'react-redux';
 import { AgentTools } from './agent_tools';
@@ -21,6 +21,25 @@ jest.mock('../../../../services', () => {
     ...mockCoreServices,
   };
 });
+
+// Mock SimplifiedJsonField to make onChange/onBlur testable
+let capturedOnChange: ((value: string) => void) | undefined;
+let capturedOnBlur: ((value: string) => void) | undefined;
+jest.mock(
+  '../components/simplified_json_field',
+  () => ({
+    SimplifiedJsonField: (props: any) => {
+      capturedOnChange = props.onChange;
+      capturedOnBlur = props.onBlur;
+      return (
+        <div data-testid="simplifiedJsonField">
+          <span>{props.label}</span>
+          {props.isInvalid && <span data-testid="fallbackQueryError">{props.error}</span>}
+        </div>
+      );
+    },
+  })
+);
 
 // Setup mock store
 const mockStore = configureStore([]);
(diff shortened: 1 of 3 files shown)


owaiskazi19 (approved):
Thanks for adding the validation as well

public/pages/workflow_detail/agentic_search/configure_flow/agent_tools.test.tsx
```diff
@@ -107,6 +107,9 @@ describe('AgentTools', () => {
     // Check that the generation type field is rendered
     expect(screen.getByTestId('generationTypeField')).toBeInTheDocument();
     expect(screen.getByTestId('generationTypeRadioGroup')).toBeInTheDocument();
+
+    // Check that the fallback query field is rendered
+    expect(screen.getByText('Fallback query')).toBeInTheDocument();
```
owaiskazi19, line 131: Can we have some test for validating the fallback queries?

PR #917, opened 2026-09-14
Guard the Access column against a data-source staleness race

DarshitChanpura:
### Description

`getResourceSharingAvailableTypes(dataSourceId)` is called asynchronously whenever the selected data source changes, but the resolved types were stored without recording which data source they belong to. On a data-source switch, the previous data source's resolved types could still decide the Access column's visibility in `workflow_list.tsx` for the duration of the new, in-flight probe.

Found while addressing [kaituo's review](https://github.com/opensearch-project/anomaly-detection-dashboards-plugin/pull/1238) on anomaly-detection-dashboards-plugin#1238, and confirmed to be missing from every resource-sharing consumer plugin on inspection — not specific to this plugin.

### Fix

Stores resolved types alongside the data source id they belong to, and only passes the resolved types into `columns(...)` when that id still matches the currently selected `dataSourceId`; otherwise passes `[]`. Kept the guard at the call site in `workflow_list.tsx` rather than changing `columns(...)`'s exported signature.

### Category
Bug fix

### Testing
Adds direct unit coverage of the guard's logic (resolved types returned once they match the currently selected data source; empty list returned for a stale/previous data source or before any result resolves).

### Check List
- [x] New functionality includes testing


DarshitChanpura:
### Re: Code-Diff-Analyzer flag on `opensearch_dashboards.json`

The analyzer flagged the new `"securityDashboards"` entry in `optionalPlugins` per its mandatory "flag all new dependency changes" rule — this is expected, not a defect.

**For maintainer review:**
- `securityDashboards` resolves to the in-repo `security-dashboards-plugin` (relative import, e.g. `../../security-dashboards-plugin/public/types`) — not an npm package, so there is no typosquatting/namespace-hijack surface.
- It is declared in `optionalPlugins`, not `requiredPlugins`, matching the existing pattern already used for `assistantDashboards`/`dataSource`/etc. in this manifest.
- The dependency is optional at every level: an optional TypeScript field (`securityDashboards?: SecurityPluginStart`), and a runtime `if (securityDashboards)` guard in `start()`. When security-dashboards-plugin is absent or disabled, that block never runs and the resource-sharing helper fails closed to `[]` without ever calling into it — covered by a dedicated unit test (`returns [] without probing when security-dashboards-plugin is not installed`).
- The same dependency + pattern is added identically across the sibling PRs below, so a single review here covers the shape of the change everywhere it appears.

Once reviewed, the standard unblock is: add the `skip-diff-analyzer` label, then re-run the failed check.

DarshitChanpura:
Closing this draft PR. The premise was incorrect: this plugin's per-data-source resource-sharing check was already correct, and gating it further on security-dashboards-plugin's local resource-sharing flag introduced a regression (the Access column would incorrectly disappear when local is disabled but the selected data source has sharing enabled).

The real defect (reported by kaituo on anomaly-detection-dashboards-plugin#1238) was in security-dashboards-plugin itself: its Share-button DOM-marker SPI watcher only started when the local cluster had resource sharing enabled. That is fixed directly in opensearch-project/security-dashboards-plugin#2525 (the SPI watcher now always starts), which requires no changes in this plugin.

DarshitChanpura:
Reopened with a different fix than what the earlier bot analysis/review comments above refer to (those were against the now-reverted local-SPI-gate approach on commit ae3f2a7).

This PR now adds a data-source staleness guard to the Access column's resource-sharing availability check in `workflow_list.tsx` — see the updated description above. No `securityDashboards` optional dependency or manifest changes are part of this PR.

ohltyler (approved):

PR #921, opened 2026-09-17
Fix the broken model documentation links in the Set up models step

### LINKED ISSUE #11101: [BUG] Broken link in "Connect to an externally hosted model" text
**Describe the bug**

There is a broken link in the UI text that reads: "Connect to an externally hosted model and make it available in your OpenSearch cluster. Learn more"

The "Learn more" link appears to be broken and does not direct users to the correct documentation or resource.

**Expected behavior**

The "Learn more" link should direct users to the appropriate documentation or resource for connecting to externally hosted models in OpenSearch.

**To Reproduce**

1. Navigate to the relevant section in OpenSearch Dashboards where this message appears
2. Click on the "Learn more" link
3. Observe that it either leads to a 404 page or an incorrect URL

**Additional context**

This appears to be related to the externally hosted models feature in OpenSearch Dashboards. The link should point to the correct documentation resource.

Please ensure the link is corrected to point to the appropriate OpenSearch documentation or support resource.

gohri81-tech:
### Description

Fixes the broken `Learn more` link reported in opensearch-project/OpenSearch-Dashboards#11101 ("Connect to an externally hosted model and make it available in your OpenSearch cluster. Learn more", in the `Set up models` step of the AI Search Flows tab).

That link comes from `ML_REMOTE_MODEL_LINK` in `common/constants.ts`, which is used by `get_started_accordion.tsx`, `intro_flyout.tsx`, `model_field.tsx`, `ml_processor_inputs.tsx` and `quick_configure_modal.tsx`. It, and `ML_INTERFACE_LINK` next to it, were moved to the `docs.opensearch.org` host but kept the `/docs` path prefix of the old `opensearch.org/docs/...` URLs:

```
https://docs.opensearch.org/docs/latest/ml-commons-plugin/remote-models/supported-connectors/
```

On `docs.opensearch.org` the version comes straight after the host, so the `/docs` segment has to go. Every other link in the same block still uses the `opensearch.org/docs/latest/...` form, which is why only these two are broken. This removes the duplicated segment and leaves the rest of both paths alone:

```
https://docs.opensearch.org/latest/ml-commons-plugin/remote-models/supported-connectors/
https://docs.opensearch.org/latest/ml-commons-plugin/api/model-apis/register-model/#the-interface-parameter
```

If you would rather have these consistent with their neighbours instead, the equivalent change is `https://opensearch.org/docs/latest/...` for both. Happy to switch.

There is one more occurrence of the same pattern in a code comment, `public/utils/config_to_template_utils.ts` line 171, which I left out to keep this to the user-facing links. Say the word and I will fix it here too.

### Issues Resolved

Closes opensearch-project/OpenSearch-Dashboards#11101

### Check List

- [x] New functionality has been documented: not applicable, this is a URL change

I cannot open the documentation site or run Dashboards from here, so I reasoned from the URL shapes already in this file rather than by clicking through; please sanity check the two targets before merging.


gohri81-tech:
A gentle nudge on this one when a maintainer has a moment. It drops the duplicated `/docs` segment from the two `docs.opensearch.org` links in `common/constants.ts`, which is what makes the "Learn more" link in the Set up models step 404 today.

As mentioned above I could not open the documentation site from where I was working, so a quick sanity check of the two target URLs before merging would be welcome. I am also happy to fix the same pattern in the code comment in `public/utils/config_to_template_utils.ts` if you would like that in here. Thank you.

gohri81-tech:
Pushed the follow-up I offered above: the same duplicated /docs segment in the code comment in public/utils/config_to_template_utils.ts is now fixed too, so all three links in this repository point at docs.opensearch.org/latest/... . The PR is still URL-only, no behaviour change.

Still looking for a maintainer review when someone has a moment.

